PR #33, opened 2023-12-07
chore(cardano-node-follower): Crate API design

### LINKED ISSUE #12: 🛠️ [TASK] : Define the initial API surface for the crate (See Description for high level overview of API requirements).
### Summary

Defines the high level API for all planned api functions

### Description

Functions should effectively be No-ops, and not have any real functionality yet.
It just defines the configuration structures and the API surface itself.
At the end of this the API should be fully defined and properly documented and Rust docs being produced.
Review of the produced rust-docs will be necessary to sign off on this task.

This task is better as a collaborative effort between 2 developers, with a review by the whole team following that.
API to follow architectural guidance.

FelipeRosa:
This PR adds some code to the `cardano-chain-follower` that roughly defines the crate's API. I also added some examples, which don't work because there's no implementation yet.

Closes #12.


hermes/crates/cardano-chain-follower/src/lib.rs
```diff
@@ -1,6 +1,213 @@
...
+    /// Pallas node-to-client client.
+    N2c(NodeClient),
+}
+
+/// Enum of chain updates received by the follower.
+pub enum ChainUpdate {
+    /// New block inserted on chain.
+    Block(MultiEraBlockData),
+    /// Chain rollback to the given point.
+    Rollback(Point),
```
Mr-Leshiy, line 125: Not sure about `Rollback(Point)` it is possible to change it to `Rollback(MultiEraBlockData)` ?

hermes/crates/cardano-chain-follower/src/lib.rs
```diff
@@ -1,6 +1,213 @@
...
+    /// not found on the chain.
+    ///
+    /// # Arguments
+    ///
+    /// * `at`: Point at which to set the read-pointer.
+    ///
+    /// # Errors
+    ///
+    /// Returns Err if something went wrong while communicating with the producer.
+    pub async fn set_read_pointer<P>(&mut self, _at: P) -> Result<Option<Point>>
```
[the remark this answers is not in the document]
FelipeRosa: I think that would make the crate's API feel weird for someone who's using it because of the mixed functionalities. e.g. What if I want to just set the read pointer?

hermes/crates/cardano-chain-follower/src/lib.rs
FelipeRosa: Yeah that would be possible but we need to figure out fetching blocks for both node-to-node and node-to-client connections (because the node just returns the point in case of a rollback).

Mr-Leshiy (approved):
LGTM

stevenj (approved):
LGTM

PR #55, opened 2024-01-08
feat: CDDL module structure

### LINKED ISSUE #54: Create multiple parser modules, so that we can have a rfc8610 parser, an rfc8610+rfc9165 parser, an rfc8610+rfc9165+modules parser, and a parser that's just used for integration tests which includes the test grammar, and all other grammars.  Add an enum so when the parser is called, it can use the correct grammar.  Modules to be named

apskhem:
# Description

To enhance our parsing capabilities, we aim to implement different parsers to handle various scenarios. Specifically, we plan to develop parsers for RFC8610, RFC8610 combined with RFC9165, and RFC8610 combined with RFC9165 along with additional modules, and a dedicated parser for integration tests. To facilitate this, we propose the addition of an enumeration that allows the parser to select the appropriate grammar based on the specified requirements. Additionally, we will assign names to the modules to ensure clarity and ease of reference.

## Related Issue(s)

> e.g., Closes #54 

## Description of Changes

* Added new structure for supporting parser modules
* Changed `crate::parse_cddl` to accept two arguments, one is the text input, and another is extension to be parsed
* Changed `CDDLError` from a struct to enum to support multiple error variants. And also moved the error to the dedicated module.


hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,20 +7,53 @@
...
-#[grammar = "grammar/cddl_test.pest"] // Ideally this would only be used in tests.
-pub struct CDDLParser;
+pub mod rfc_8610 {
+    pub use pest::Parser;
+
+    #[derive(pest_derive::Parser)]
+    #[grammar = "grammar/cddl.pest"]
+    pub struct RFC8610Parser;
+}
+
```
stevenj, line 22: Add the other `.pest` files that are planned, and just have them with a comment at the top explaining what they will contain.

Then add the other 2 parser modules we plan here, so that in future all we need to do is add the grammar and tests and its done.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -45,41 +78,61 @@ pub const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
...
-        Err(e) => {
-            println!("{e:?}");
-            println!("{e}");
-            return Err(Box::new(CDDLError::from(e)));
+pub fn parse_cddl(input: &str, extension: &Extension) -> Result<(), Box<CDDLError>> {
+    match extension {
+        Extension::RFC8610Parser => {
+            let result = rfc_8610::RFC8610Parser::parse(rfc_8610::Rule::cddl, input);
+
+            match result {
```
stevenj, line 93: This is common logic,  just have the `rfc_8610::...` use `map_err` https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err to convert the error to the enum error type.
Then the logic is cleaner.
You can then use:
```
let result = match extension {
...
}
```

and then just print the error in one after the match.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -45,41 +78,61 @@ pub const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
...
+                Err(e) => {
+                    println!("{e:?}");
+                    println!("{e}");
+                    Err(Box::new(CDDLError::from(CDDLErrorType::RFC8610(e))))
+                },
+            }
         },
-    }
+        Extension::RFC9615Parser => {
+            unimplemented!()
```
stevenj, line 103: these can be implemented if we add the placeholder `.pest` files and declare the modules as mentioned above.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,20 +7,53 @@
...
+pub enum Extension {
+    /// RFC8610 ONLY limited parser.
+    RFC8610Parser,
+    /// RFC8610 and RFC9615 limited parser.
+    RFC9615Parser,
+    /// RFC8610, RFC9615, and CDDL modules.
+    CDDLParser,
+    /// Same as CDDLParser but includes the `cddl_test.pest` file for integration test
+    /// usage, mainly for development testing.
+    CDDLTestParser,
```
stevenj, line 43: Because we can call the `cddl_test` parser directly in the unit tests, we can exclude it from this enum, and also the error type.  That way the public API only has the valid parsers, not out test parser which is only used for the Unit tests.

stevenj (changes requested):

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,23 +7,80 @@
...
-#[grammar = "grammar/cddl.pest"]
-#[grammar = "grammar/cddl_test.pest"] // Ideally this would only be used in tests.
-pub struct CDDLParser;
+pub mod rfc_8610 {
+    pub use pest::Parser;
+
+    #[derive(pest_derive::Parser)]
+    // #[grammar = "grammar/rfc_8610.pest"]
+    // TODO: we will implement to support those specifications later
+    #[grammar = "grammar/cddl.pest"]
```
stevenj, line 21: rename this to rfc_8610.pest

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,23 +7,80 @@
...
+    pub struct RFC8610Parser;
+}
+
+pub mod rfc_9615 {
+    pub use pest::Parser;
+
+    #[derive(pest_derive::Parser)]
+    // #[grammar = "grammar/rfc_8610.pest"]
+    // #[grammar = "grammar/rfc_9615.pest"]
```
stevenj, line 30: Add this `.pest` file, but just put a comment in it and no grammar.
That way it can be properly referenced here already.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,23 +7,80 @@
...
+    pub struct RFC8610Parser;
+}
+
+pub mod cddl {
+    pub use pest::Parser;
+
+    #[derive(pest_derive::Parser)]
+    // #[grammar = "grammar/rfc_8610.pest"]
+    // #[grammar = "grammar/rfc_9615.pest"]
+    // TODO: we will implement to support those specifications later
```
stevenj, line 42: add a `cddl_modules.pest` file here, and like above just put a comment in it as a placeholder.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -7,23 +7,80 @@
...
+    #[grammar = "grammar/cddl.pest"]
+    pub struct RFC8610Parser;
+}
+
+pub mod cddl_test {
+    pub use pest::Parser;
+
+    // Parser with DEBUG rules. These rules are only used in tests.
+    #[derive(pest_derive::Parser)]
+    // #[grammar = "grammar/rfc_8610.pest"]
```
stevenj, line 52: You can already include all the new `.pest` files here.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -45,35 +102,50 @@ pub const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
...
+    })
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::{parse_cddl, POSTLUDE};
+    use crate::*;
+
+    // CDDL Standard Postlude - read from an external file
+    const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
```
stevenj, line 144: I think we should promote this outside the test module, up at the top of the file after the parser modules are defined.
I also think in the parser now, we should get `input` and append the `POSTLUDE` to it in `parse_cddl`.  This is because every CDDL file should have the `POSTLUDE` added to it according to the spec.

This test then can just try and parse an empty string which is equivalent to just parsing the `POSTLUDE`

stevenj (changes requested):

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -45,35 +107,52 @@ pub const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
...
-    match result {
-        Ok(c) => println!("{c:?}"),
-        Err(e) => {
-            println!("{e:?}");
-            println!("{e}");
-            return Err(Box::new(CDDLError::from(e)));
+pub fn parse_cddl<'a>(
+    input: &'a str, extension: &Extension,
+) -> Result<Box<AST<'a>>, Box<CDDLError>> {
+    let result = match extension {
```
stevenj, line 120: `POSTLUDE` needs to be added to `input` here before its parsed.

hermes/crates/cbork/cddl-parser/src/lib.rs
```diff
@@ -45,35 +107,52 @@ pub const POSTLUDE: &str = include_str!("grammar/postlude.cddl");
...
 
 #[cfg(test)]
 mod tests {
-    use crate::{parse_cddl, POSTLUDE};
+    use crate::*;
 
     #[test]
     fn it_works() {
-        let result = parse_cddl(POSTLUDE);
+        let result = parse_cddl(POSTLUDE, &Extension::CDDLParser);
```
stevenj, line 155: When `POSTLUDE` is automatically included by the parser, this test can just check an empty string.

stevenj (changes requested):
This should be the last thing.

stevenj (approved):
LGTM

Issue #70, opened 2024-01-25
🛠️ [TASK] : Document Hermes application packaging requirements


saibatizoku:
### Summary

Describe Hermes application packaging requirements using HDF5

## Applications

- Metadata Describing the Application.
  - Name
  - Author
  - Icons
  - Description
  - Configuration
  - etc
- Static Files which can be served AS-IS to a local host over HTTP.
- Static Data which can be used by Webasm Modules.
- Webasm Modules to Implement the Dynamic Functionality of the application.
- Signature: All Data in the application needs to be able to be signed for authentication purposes.

Applications are read only.  The data within the Application can not be updated as it is run.

### Static Files
A large part of any web application is a collection of static assets such as:

  - Images.
  - HTML / CSS.
  - Javascript / Webasm to execute in the browser.
  - This is such a common use case that the Hermes should have the in-built capability to serve these files from a known SUB DOMAIN and path used by the application.

The Webasm modules would ALSO be able to read this data.

## Static Files
This is data which the Dynamic Functionality provided by the WebAsm Modules needs in order to do its job.  An example could be javascript files which are templated and need to be processed before serving.  The type and kind of these files is unbounded, but necessary for lots of application purposes.

These files are NOT directly accessible over HTTP from the Application but can be read from the Webasm.

## Webasm Modules

Webasm Modules are a kind of Mini Application and they will need:
- Metadata describing the module.
  - Name
  - Author
  - URL of the Source
  - Configuration
  - etc

- Static Data that can ONLY be used by the Module itself.
  - Examples could be tabular data.

- The Webasm binary code itself.
- Signature: All Data in the webasm module needs to be able to be signed for authentication purposes.

PR #71, opened 2024-01-25
docs: Adds notes on Hermes packaging requirements

saibatizoku:
# Description

Adds documentation requirements for packaging Hermes Applications.

## Related Issue(s)

#70 

## Description of Changes

See the generated docs [here](https://input-output-hk.github.io/hermes/branch/feat_hermes_packaging_docs/architecture/08_concepts/hermes_packaging_requirements/overview/).

- [x] Overview
- [x] Packaging A Hermes Application
- [x] Defining Application Metadata
- [ ] Static Assets
- [ ] Application Data
- [ ] Application WASM Modules
- [ ] Application Signatures


Mr-Leshiy (dismissed):
Looks awesome !

But I have a question about private and public data, maybe it make sense to rename directories to the `var/public` and `var/private` instead of `var/data`, `var/www` ?
So it will make it more clear and from the packaging perspective we will don't care what exact data should be there, we are just managing the scope to it.

cc @stevenj

saibatizoku:
> Looks awesome !
> 
> But I have a question about private and public data, maybe it make sense to rename directories to the `var/public` and `var/private` instead of `var/data`, `var/www` ? So it will make it more clear and from the packaging perspective we will don't care what exact data should be there, we are just managing the scope to it.
> 
> cc @stevenj

Thanks! I'm open to renaming the paths, and I considered a similar approach to what you are proposing, but I think that `public` and `private` are terms that could lead to confusion, as they have different meanings depending on the context. 

In this case, the context is using HDF5 files as a file-system, so using `/var/data` and `/var/www`, is similar to the convention in POSIX-based file-systems that should be familiar to most developers.

So, in this case, I would opt for keeping the names as they are, or to find a middle ground.

FelipeRosa (approved):
LGTM

bkioshn (commented):
Looks good, the document is clear and understandable 🥳

PR #73, opened 2024-01-30
feat: Wasm module implementation 

### LINKED ISSUE #65: 🛠️ [TASK] : WASM runtime engine setup
### Summary

Prepared an initial WASM runtime engine setup

### Description

Need to implement a wasm runtime setup for running and execution wasm modules, as a separate library crate.
use https://github.com/bytecodealliance/wasmtime as a WASM runtime engine.

Mr-Leshiy:
# Description

An initial WASM module implementation

## Related Issue(s)

Closes #65 

## Description of Changes

- Added a new `wasm` Rust module where going to put all WASM specific code
- Added new `Module` struct which is a wrapper over `wasmtime::Module`, `wasmtime::InstancePre` etc.. It initialises a module from provided wasm module bytes, context, imports.
- Added a `call_func` function, to call some `export` functions of the WASM module from the Rust. Also this function preserves an immutable WASM. state as it was described in the ticket.
- Added a `Context` struct which represents a host context to the WASM module, and will be a part of the WASM Store.


saibatizoku (approved):
LGTM.

Mr-Leshiy:
depends on https://github.com/input-output-hk/catalyst-ci/pull/180

hermes/bin/src/wasm/module.rs
```diff
@@ -0,0 +1,140 @@
...
+        fn link(&self, linker: &mut WasmLinker<Context>) -> Result<(), Box<dyn Error>> {
+            linker.func_wrap("", "hello", || {
+                println!("hello");
+            })?;
+            Ok(())
+        }
+    }
+
+    #[test]
+    /// Tests that after instantiation of `Module` it's state does not change after each
```
saibatizoku, line 113:
```suggestion
    /// Tests that after instantiation of `Module` its state does not change after each
```

saibatizoku (dismissed):
LGTM

saibatizoku (dismissed):
:+1:

hermes/bin/src/wasm/module.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///  - `wasmtime::Error`: WASM call error
+    #[allow(dead_code)]
+    pub(crate) fn new(
+        context: Context, module_bytes: &[u8], imports: &[Box<dyn LinkImport<Context>>],
+    ) -> Result<Self, Box<dyn Error>> {
+        let mut config = WasmConfig::new();
+        config.wasm_component_model(true);
+        config.consume_fuel(false);
+
+        let engine = WasmEngine::new(&config)?;
```
stevenj, line 59: There should only be one engine, for all modules.
You can make an `engine` crate which creates a new engine, and then we just keep re-using it.
Pass the engine into the `new` as a parameter.
Still save it in the Module struct, just in case in future we do something like have a different engine for each application.

hermes/bin/src/wasm/module.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl Module {
+    /// Instantiate WASM module
+    ///
+    /// # Errors
+    ///  - `wasmtime::Error`: WASM call error
+    #[allow(dead_code)]
+    pub(crate) fn new(
+        context: Context, module_bytes: &[u8], imports: &[Box<dyn LinkImport<Context>>],
```
stevenj, line 53: Context should be created internally to `new`, and managed by this crate.
So we don't need it here.
What we do need however is an `app_name: string` which would be used when we make a new context for this wasm module.

hermes/bin/src/wasm/module.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let engine = WasmEngine::new(&config)?;
+
+        let module = WasmModule::new(&engine, module_bytes)?;
+
+        let mut linker = WasmLinker::new(&engine);
+        for import in imports {
+            import.link(&mut linker)?;
+        }
+        let instance = linker.instantiate_pre(&module)?;
```
stevenj, line 67: call this a `pre_instance` both here and in the struct that its not a full instance yet.

hermes/bin/src/wasm/module.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[allow(dead_code)]
+    pub(crate) fn call_func<Args, Ret>(
+        &mut self, name: &str, args: Args,
+    ) -> Result<Ret, Box<dyn Error>>
+    where
+        Args: WasmParams,
+        Ret: WasmResults,
+    {
+        let mut store = WasmStore::new(&self.engine, self.context.clone());
+        let instantiated_instance = self.instance.instantiate(&mut store)?;
```
stevenj, line 91: I would just call this `instance` and the other `pre_instance`.

stevenj (changes requested):

hermes/bin/src/wasm/context.rs
```diff
@@ -0,0 +1,6 @@
+//! WASM host context implementation.
+
+/// A WASM host context structure, which is intended to be passed to the `wasmtime::Store`
+/// during the WASM state initialization process.
+#[derive(Clone)]
+pub(crate) struct Context;
```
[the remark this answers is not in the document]
stevenj: You can make a "general" instance.  and make `new` which creates it.
Then you can have a method on it called `use_for(event:string)` which increments the Execution Count in the base instance, then clones the base instance and sets the `Event name` field to the parameter.
This then returns a "usable" context which we would attach to the store.

The reason I want these two extra parameters, is so when we call functions in the runtime, we can check is the function being called by the same "event" within the module, and also, is it the same "execution".  it would allow the extension to control if its state can be shared between executions or is bounded by a particular execution.

The choice and how these fields get used is up to the runtime extension, but with these 4, we have pretty fine granularity over managing the external state accurately.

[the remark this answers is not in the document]
stevenj: I would use this for the execution counter:
https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html
As it will be thread safe,  and should be high performance.

stevenj (changes requested):

stevenj (approved):

PR #80, opened 2024-02-02
feat: genesis values helper

cong-or:
Helper to obtain genesis values which are required for` block.epoch(genesis: &GenesisValues)` which in my use case; is consumed from the gateway.

Gateway->Follower->Pallas.

When consuming via catalyst gateway and importing pallas and the follower to satisfy types, conflicts emerge. It is cleaner to just import the follower and add this helper method.


FelipeRosa (approved):
LGTM

Mr-Leshiy (approved):
LGTM

PR #118, opened 2024-02-07
feat(wasm): C stub

### LINKED ISSUE #117: 🛠️ [TASK] : `C` stub wasm module
### Summary

`C` stub wasm module

### Description

Implement all defined in hermes `.wit` file exports in `C` and prepare a builder to build wasm module.

The task here is to make a minimal C wasm component module for our `hermes` world, defined by our wasm `.wit` files.

The wit-bindgen repo has an example for how to build such a module with C:
https://github.com/bytecodealliance/wit-bindgen?tab=readme-ov-file#guest-cc

We need an earthly builder that sets up the required C compiler and tooling.
Generates the necessary files using wit-bindgen (there is an integration test already which does this, this code can be moved for the C integration test to this module.)

Then compile the C stub to a .wasm file and save it as an artifact, so it can be used by our other builders.

There is a branch doing something similar (but not exactly the same) for go here:
https://github.com/input-output-hk/hermes/tree/test/wasm-go

The stubs have no implementation, they do nothing.
The only code they may have is to return data if the exported function signature requires a return, in which case we always return a failed or false status from the stub, never a positive or successful status.

The stubs must not call any available imports, as they need to be as minimal as possible, and have no side effects.

apskhem:
# Description

Implement all events defined in Hermes `.wit` file exports in C and prepare a builder to build the wasm module.

## Related Issue(s)

Closes #117 

## Description of Changes

* Implemented stuff in C that needed to be compiled as stub functions
* Added an Earthfile for automated compilation into `.wasm` component


minikin (commented):
I'm wondering if C has some modern linters/formaters?

wasm/event-stubs/Earthfile
```diff
@@ -0,0 +1,29 @@
...
+
+build-c-bindings:
+    FROM ../wasi-hermes-component-adapter+builder
+
+    # FIXME: right now we need to remove `--no-object-file` flag to make it runs with our bench, should we add this back?
+    RUN wit-bindgen c --autodrop-borrows yes ../wasi/wit
+
+    SAVE ARTIFACT ./
+
+build-c-stub:
```
bkioshn, line 13: Add general comment on what this target do (if possible)

wasm/event-stubs/Earthfile
```diff
@@ -0,0 +1,29 @@
...
+    FROM ghcr.io/webassembly/wasi-sdk:wasi-sdk-21
+    WORKDIR /repo
+    COPY . .
+    COPY +build-c-bindings/ ./
+
+    RUN $CC hermes.h hermes_component_type.o my-component.c -o out.wasm -mexec-model=reactor $CFLAGS
+
+    SAVE ARTIFACT ./out.wasm
+
+build-c-component:
```
bkioshn, line 23: ditto

wasm/event-stubs/Earthfile
```diff
@@ -0,0 +1,29 @@
+VERSION 0.7
+
+# cspell: words wasi autodrop webassembly WORKDIR CFLAGS mexec
```
bkioshn, line 3: `wasi` `webassembly` `WORKDIR` seem to be common word using across the project, might move them to the dict

apskhem:
> I'm wondering if C has some modern linters/formaters?

Currently, we don't have any checks in place. For now, the code just needs to be simple and work. If we decide to add those checks later on, we can update and improve the code at that time.

bkioshn (commented):
Looks good to me 🥳

wasm/event-stubs/Earthfile
```diff
@@ -0,0 +1,32 @@
...
+    SAVE ARTIFACT ./
+
+# build-c-stub - compiles the generated C files into .wasm module
+build-c-stub:
+    FROM ghcr.io/webassembly/wasi-sdk:wasi-sdk-21
+    WORKDIR /repo
+    COPY . .
+    COPY +build-c-bindings/ ./
+
+    RUN $CC hermes.h hermes_component_type.o my-component.c -o out.wasm -mexec-model=reactor $CFLAGS
```
Mr-Leshiy, line 21: Better place what exact compiler you are using right here instead of `$CC`.
Because it is possible that somehow `CC` env var would be changed and this line will not work anymore.

wasm/event-stubs/Earthfile
```diff
@@ -0,0 +1,32 @@
...
+
+    SAVE ARTIFACT ./out.wasm
+
+# build-c-component - turns a C .wasm module into a component ready to be used with Hermes
+build-c-component:
```
Mr-Leshiy, line 26: Are you sure that this target is the correct one ?
How you are propagating `out.wasm` file which is generated inside `build-c-stub` ?

wasm/event-stubs/my-component.c
```diff
@@ -0,0 +1,42 @@
+#include "hermes.h"
+
+// cspell: words cardano
```
Mr-Leshiy, line 3: Better move this word into the `project.dic` file, because I suppose it would be widely used.

Mr-Leshiy (changes requested):

wasm/event-stubs/Earthfile
Mr-Leshiy: And the same for the `$CFLAGS` better put flags explicitly

wasm/event-stubs/Earthfile
apskhem: I missed that. It needs to copy the result from `build-c-stub`, and processes it. Just wonder why the CI didn't catch any errors.

Mr-Leshiy (approved):
LGTM

Issue #125, opened 2024-02-13
🛠️ [TASK] : Hermes event queue 


Mr-Leshiy:
### Summary

Add a Hermes event queue implementation

### Description

Add a hermes event queue implementation as a multi producer single consumer structure ([spec](https://input-output-hk.github.io/hermes/architecture/08_concepts/hermes_runtime_engine/h_events_queue/)).

PR #135, opened 2024-02-16
feat: Hermes event queue and Hermes reactor

Mr-Leshiy:
# Description

- Added a `HermesEventQueue` implementation as a mpsc FIFO queue.  This struct is s singleton object which is stores as a static `OnceCell` variable.
- Added a `HermesReactor` struct, which should a general manager which orchestrates all hermes execution like: running and stepping runtime extensions; initialising and holding an event queue instance and manages event loop thread.
- Refactored and modified state management.

## Related Issue(s)

#125


hermes/bin/src/event/queue.rs
```diff
@@ -0,0 +1,164 @@
...
+    AnotherEventExecutionLoop,
+}
+
+/// Hermes event queue.
+pub(crate) struct HermesEventQueue {
+    /// Hermes event queue sender
+    sender: Sender<HermesEvent>,
+    /// Hermes event queue receiver
+    receiver: Mutex<Receiver<HermesEvent>>,
+
```
FelipeRosa, line 46: super nitpick: empty line

hermes/bin/src/runtime_extensions/hermes/binary/mod.rs
```diff
@@ -1,14 +1,14 @@
...
 
 /// State
 pub(crate) struct State {}
 
 impl Stateful for State {
-    fn new(_ctx: &Context) -> Self {
+    fn new() -> Self {
         State {}
     }
 }
```
FelipeRosa, line 14: I don't see the need for this trait because I don't think we have any generic/dynamic code that would require this. 

Each runtime extension `State` type can implement a function for creating an instance of it and we can just call it when creating the aggregate state (the one containing all extensions states) since it's all known at compile-time.

WDYT?

hermes/bin/src/event/queue.rs
```diff
@@ -0,0 +1,164 @@
...
+                            module_id.clone(),
+                            state,
+                            event.payload(),
+                            module,
+                        )?;
+                    }
+                },
+            }
+            Ok(())
+        };
```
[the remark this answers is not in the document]
Mr-Leshiy: Yea, make sense.

hermes/bin/src/event/queue.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Target app not found.
+    #[error("Target app not found.")]
+    AppNotFound,
+
+    /// Target module not found.
+    #[error("Target module not found.")]
+    ModuleNotFound,
+
+    /// Failed to add event into the event queue. Event queue is closed.
+    #[error("Failed to add event into the event queue. Event queue is closed.")]
```
bkioshn, line 31: Just curious, Can other issues prevent events from being added to the event queue?

Mr-Leshiy: Do you mean like `AnotherEventExecutionLoop` or `ModuleNotFound`/`AppNotFound` ?
No, they are not preventing and not locking for adding new events

hermes/bin/src/runtime_extensions/hermes/binary/mod.rs
Mr-Leshiy: yea, you are right at this moment it is not used as a trait (as a trait boundary and as a part of dynamic code).
But potentially we could need it later (or not, will see 😊).
Initially it was added as a additional reminder that for each runtime_extensions it is mandatory to add a `State` struct even if it is not contain anything.

Mr-Leshiy: For now I would leave it as it is, it could be removed at any time later when we will have more mature state of Hermes.

hermes/bin/src/event/queue.rs
bkioshn: I mean the problem with the queue itself.

hermes/bin/src/runtime_extensions/hermes/binary/mod.rs
stevenj: What might be cleaner however, is to just turn this into traditional functions.
As its probably always going to be necessary for any complex state to be fully managed inside the runtime module themselves, perhaps this is just cleaner if we have:

```rust
  fn new_context(ctx &context) {
      wasi::new_context(ctx);
      hermes::new_context(ctx);
  }

  // wasi module then has
  fn new_context(ctx &context) {
      cli::new_context(ctx);
      clocks::new_context(ctx);
      filesystem::new_context(ctx);
      http::new_context(ctx);
      io::new_context(ctx);
      random::new_context(ctx);
  }
   
 // hermes module has
 fn new_context(ctx &context) {
   binary::new_context(ctx);
   cardano::new_context(ctx);;
   // and so on...
 }

 // and so on....
```

This will make it explicit that we are not dealing with an actual memory allocation in the state, force each runtime extension to manage its own state given a particular context which it needs to do anyway and remove ambiguity.

This only gets called once when a new context is actually made (new app, module and execution count for the module changes).

stevenj (changes requested):

hermes/bin/src/event/queue.rs
```diff
@@ -0,0 +1,194 @@
...
+    #[error("Trying to execute one more event execution loop. It is allowed to run only one execution loop in a time.")]
+    AnotherEventExecutionLoop,
+}
+
+/// Hermes event queue.
+pub(crate) struct HermesEventQueue {
+    /// Hermes event queue sender
+    sender: Sender<HermesEvent>,
+    /// Hermes event queue receiver
+    receiver: Mutex<Receiver<HermesEvent>>,
```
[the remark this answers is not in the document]
Mr-Leshiy: Technically yes, but `Receiver` is a part of the HermesQueue itself which could be passed through different threads, because we need to pass `Sender` to another runtime extensions.

hermes/bin/src/reactor.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl HermesReactor {
+    /// Create a new Hermes Reactor
+    #[allow(dead_code)]
+    pub(crate) fn new(apps: Vec<HermesApp>) -> Self {
+        let event_queue = HermesEventQueue::new().into();
+
+        let state = State::new().into();
+        let indexed_apps = apps
```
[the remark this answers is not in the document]
Mr-Leshiy: Yea, I have this idea in mind, so supposing to add a `DashMap` for it and additional functionality how to dynamically add and remove apps.

hermes/bin/src/reactor.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// Run Hermes.
+    ///
+    /// # Note:
+    /// This is a blocking call util all tasks are finished.
+    #[allow(dead_code)]
+    pub(crate) fn run(self) -> anyhow::Result<()> {
+        // Emits init event
+        self.state
```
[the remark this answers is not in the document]
Mr-Leshiy: yea, agree, but I would add this functionality later.

hermes/bin/src/runtime_extensions/hermes/binary/mod.rs
Mr-Leshiy: > Why did we remove the &Context? 
this Context (renamed to the `HermesRuntimeContext` inside `runtime_state.rs`) as a part of the `HermesRuntimeState`. `HermesRuntimeState` - actually initialised for every WASM execution and passed as a state to WASM.
So when we are talking about host implementation of the runtime extensions, they are using this `HermesRuntimeState` and could get a proper context how they are running.
But the runtime extension itself when we are defining it and initialising dont know and should not know about the this Context, because they are not running as a part of some WASM execution. They are separate.
```Rust
pub(crate) struct HermesRuntimeState {
    /// Runtime extensions state
    pub(crate) state: Arc<State>,
    /// Runtime context.
    pub(crate) _ctx: HermesRuntimeContext,
    /// Hermes event queue.
    pub(crate) _event_queue: Arc<HermesEventQueue>,
}
```

Mr-Leshiy: @FelipeRosa from the latest discussion with @stevenj decided to remove this trait

hermes/bin/src/event/queue.rs
```diff
@@ -5,16 +5,21 @@ use std::{
...
+
 use super::{HermesEvent, HermesEventPayload, TargetApp, TargetModule};
 use crate::{
     app::{HermesAppName, IndexedApps},
     runtime_context::HermesRuntimeContext,
     wasm::module::{Module, ModuleId},
 };
 
+/// Singleton instance of the Hermes event queue.
+static EVENT_QUEUE_INTANCE: OnceCell<HermesEventQueue> = OnceCell::new();
```
FelipeRosa: Suggestion: if we use `Lazy` (since `LazyLock` is still unstable) we can make the `get_instance` method to return the queue instance directly instead of a `Result`. That way every runtime extension that wishes to use the event queue won't need to check for not initialized errors.

Mr-Leshiy: But the key problem here, that you cannot initialise it in a random place, for initialisation you need to provide a list of apps, which can not be done by the runtime extensions

FelipeRosa: Hum... got it.

FelipeRosa (commented):
LGTM. I'm just not sure about if are we going to include some mechanism for initialializing each runtime extension in this PR as mentioned by Steven?

hermes/bin/src/event/queue.rs
```diff
@@ -0,0 +1,178 @@
...
+        let event_loop = thread::spawn(move || event_execution_loop(&indexed_apps, receiver));
+
+        Ok(event_loop)
+    }
+
+    /// Get the singleton instance of the `HermesEventQueue`.
+    ///
+    /// # Errors:
+    /// - `Error::NotInitialized`
+    pub(crate) fn get_instance() -> anyhow::Result<&'static HermesEventQueue> {
```
stevenj, line 79: Don't expose OnceCell initialized data outside the module its created in.
Just make a single `add_event` function outside the `impl` and make it.
see suggested changes in PR to your branch.

hermes/bin/src/runtime_extensions/hermes/mod.rs
```diff
@@ -13,47 +11,3 @@ pub(crate) mod json;
 pub(crate) mod kv_store;
 pub(crate) mod localtime;
 pub(crate) mod logging;
-
-/// Hermes extensions state
-pub(crate) struct State {
```
stevenj, line 18: It is important to let the runtime extensions know of a new context before they receive it.
See requested changes PR for details

stevenj (changes requested):
See comments and change requested PR here:
https://github.com/input-output-hk/hermes/pull/173

hermes/bin/src/event/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// # Returns
+    ///
+    /// An `anyhow::Result` indicating the success or failure of the payload execution.
+    fn execute(&self, module: &mut ModuleInstance) -> anyhow::Result<()>;
+}
+
+/// Target Hermes app to execute the event
+pub(crate) enum TargetApp {
+    /// Execute for all available apps
+    _All,
```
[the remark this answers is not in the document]
Mr-Leshiy: It is the same case as for example with unused variable or unused function etc.
As this enum option is not initialised at any place, so compiler asks to mark it with underscore or add `allow(dead_code)`

hermes/bin/src/event/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Execute for a specific list of apps
+    List(Vec<HermesAppName>),
+}
+
+/// Target WASM module to execute the event
+pub(crate) enum TargetModule {
+    /// Execute for all available modules
+    All,
+    /// Execute for a specific list of modules
+    _List(Vec<ModuleId>),
```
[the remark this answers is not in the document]
Mr-Leshiy: same as above

hermes/bin/src/event/mod.rs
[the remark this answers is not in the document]
stevenj: Would the compiler complain anymore if there was a test that used `All` ?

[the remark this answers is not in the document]
Mr-Leshiy: Yes, it should

stevenj:
> LGTM. I'm just not sure about if are we going to include some mechanism for initialializing each runtime extension in this PR as mentioned by Steven?

Yes we did in the latest changes.

saibatizoku (approved):

FelipeRosa (approved):

saibatizoku (approved):
Thanks, this will help in getting our APIs running!

hermes/bin/src/event/queue.rs
stevenj: See: https://doc.rust-lang.org/std/sync/mpsc/struct.SendError.html
`send` can only fail if the queue is closed.  But even if it fails for another reason, the queue may as well be closed.
The queue is unbounded, so it can't be full.

stevenj (approved):
LGTM

Issue #15, opened 2023-11-30
🛠️ [TASK] : Logging utility API


stevenj:
### Summary

Define and create a Logging API for Hermes

### Description

This task will also, by necessity, involve setting up the logging infrastructure within Hermes itself.
When doing this, search for TODO with outstanding logging requirements in the code which must be implemented as part of this task.

#### Functionality to include:

1. Log messages from Hermes WASM Modules.
2. Specify Levels of logs (Debug, Info, Warn, Error).
3. Add Structured Data to Logs (JSON).
4. Display Logs with appropriate context with the running Hermes node.

#### Logging API outputs will include

* [x] #82
* [x] #83
* [x] #85
* [x] #84

PR #145, opened 2024-02-22
feat(cron api rust): Implement Cron Events Rust

### LINKED ISSUE #93: 🛠️ [TASK] : Chronological Event utility API (Cron): Hermes Implementation (Rust Module)

### LINKED ISSUE #146: 🛠️ [TASK] : Cron Event Service
### Summary

Continuation of Cron Rust Module Implementation

### Description

Continuation of Cron Rust Module Implementation

SJ Notes

The issue with namespace is that we will have multiple apps,  all of them potentially having their own crons.
A cron setup by one app, should not get sent as an event to another app, it should only target the specific app its destined for.

That should more look like :
```rust
pub(crate) struct State {
    /// The crontabs hash map.
    crontabs: DashMap<AppName,
                DashMap<CronEventTag, CronTab>>,
    cronqueue: SomeKindOfConcurrentList<(AppName, CronEventTage)> // Entries are in order from most recent to expire, to oldest to expire.
}

```

I am also not seeing a generator for the cron events themselves?

Your internal state would need to know the app name, in order to generate a cron event.
One way to do a cron task is to have an ordered list of cron events rather than a hashmap, although that could be used as an index.
The top entry in the list is the next cron to fire.
Your task sleeps on either a timeout or a trigger.
The timeout is the delay until the top item in the queue needs to be sent as an event.
When a cron is scheduled, its added to the queue at the appropriate sorted position.  Insert sort works well here,  just scan through the list until you find the position it fits between and add it.
If it was added to the head, you trigger the cron task, which will then check if it needs to immediately send it, or reschedule itself to sleep.
If its not at the head, its just added to the queue, because the task state hasn't changed.
When the task wakes up (on either timeout or trigger) it checks the head of the queue, if its ready to send, it sends it, and reschedules it back in the queue.
It then sleeps again, for the duration of the next events delay.
This is a very efficient implementation, your not polling the queue, you sleep.  The only time you check the queue is if the head changes, or you have an event to send.
If the queue is empty, you sleep forever.
You only have one entry in the queue for every event, even if they recur, because they are reinserted once they reach the head at the next time they would expire.
One thing to be aware of with this scheme, and to check for is when you have multiple crons that end at the same time, the event handler should reliably send them all, for example, if there are 3 crons that fire once a second, every second all three should fire, and be rescheduled.  If a cron is in the past, you need to send it, and reschedule it, because you may be delayed in processing under high CPU load (unlikely but it needs to be accounted for). 

You need tests to check the event queue task is being run properly.
You also need tests that add/delete/list the cron tabs from multiple threads, to ensure its all threadsafe.
You only need one cron task, because each cron entry should know what app its being generated for.  It needs to be information added to the crontab value inside the state, but never told to the applications themselves (its internal).

The cron task should be started when the state is initialized at start of the world.

The queue itself has multiple writers and readers,  its more like a list where you are mostly interested in the head than a traditional queue, thats because you need to keep entries in it sorted as they are added, and a entry could be added/removed by any add/delete of a crontab or when one expires and is either removed or re-added by thge cron task.

There are a few libraries which do this kind of thing,  but i am not in love with any of them, mostly because they are tickers, which is wildly inefficient.
If I know the next event is in 5 hours, it makes no sense to tick every 500ms to check if that 5 hours has elapsed.
But they could serve as inspiration:
https://github.com/mehcode/schedule-rs/tree/master
https://github.com/mvniekerk/tokio-cron-scheduler/tree/main
https://github.com/lholden/job_scheduler
But this is probably the cleanest/closest to what we want:
https://github.com/rs-god/rcron
I don't think any of these are exactly what we need, but the last one could form the basis of what we need adapted heavily to our use case.
We do not need a general cron API, just one that works how we want it to.
If you use this one, or any of the others as inspiration, just add a comment to the top of the file saying you used it for inspiration, just to keep the licensing/references clean.

saibatizoku:
# Description

Adds the implementation of a thread-safe Cron Event Queue as a Hermes Runtime Extension.

## Related Issue(s)

Closes #146 


hermes/bin/src/runtime_extensions/hermes/cron/mod.rs
```diff
@@ -337,8 +346,6 @@ impl Ord for CronComponent {
 
 #[cfg(test)]
 mod tests {
-    use time::Duration;
```
stevenj, line 340: The `#[allow(clippy::expect_used)]` above this line needs to be removed as does the `.expect` that its masking.
`expect` is not allowed outside of unit tests.
The node is not to panic, and `expect` is a forced panic.
far better in this case to log an error, and return that the two things are not comparable.

stevenj (changes requested):
We need tor emove all uses of `expect` outside the unit tests.
I say in the comments to log.  We haven't got a logging infrastructure in hermes yet, but you can put a line where we would log as a comment with a TODO to add logging, and tag the ticket for adding logging to hermes which is https://github.com/input-output-hk/hermes/issues/15

Also need to properly add handling for the response to the `on_cron` event.

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
```diff
@@ -0,0 +1,381 @@
...
+    InternalState::new(sender)
+});
+
+/// Name of the Application that owns the `OnCronEvent`s.
+pub type AppName = String;
+
+/// Internal State.
+pub(crate) struct InternalState {
+    /// The send events to the crontab queue.
+    pub(crate) cron_queue: CronEventQueue,
```
stevenj, line 48:
```suggestion
    cron_queue: CronEventQueue,
```

We shouldn't export the `cron_queue` directly, and it doesn't;t appear that its used anyway,

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
```diff
@@ -0,0 +1,381 @@
...
+    };
+
+    InternalState::new(sender)
+});
+
+/// Name of the Application that owns the `OnCronEvent`s.
+pub type AppName = String;
+
+/// Internal State.
+pub(crate) struct InternalState {
```
stevenj, line 46:
```suggestion
struct InternalState {
```

Shouldn't make internal state visible, and it doesn't appear that its used externally anyway.

stevenj (changes requested):
See comments for changes requested

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
saibatizoku: Yes, thanks for the catch.

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
saibatizoku: Yes, thanks for the catch.

hermes/bin/src/runtime_extensions/hermes/cron/event.rs
```diff
@@ -18,12 +26,207 @@ impl HermesEventPayload for OnCronEvent {
...
+    }
+}
+
+impl PartialEq for CronTagged {
+    fn eq(&self, other: &Self) -> bool {
+        self.tag == other.tag && self.when == other.when
+    }
+}
+
+impl Eq for CronTagged {}
```
Mr-Leshiy, line 119: Is it possible to use `derive` macro for this ?

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
```diff
@@ -0,0 +1,490 @@
...
+    event::{queue::send, HermesEvent, TargetApp, TargetModule},
+    runtime_extensions::{
+        bindings::hermes::cron::api::{CronEventTag, CronTagged, Instant},
+        hermes::cron::mkdelay_crontab,
+    },
+};
+
+/// Cron Internal State
+static CRON_INTERNAL_STATE: Lazy<InternalState> = Lazy::new(|| {
+    let sender = if let Ok(runtime) = Builder::new_current_thread().enable_all().build() {
```
Mr-Leshiy, line 26: What do you think to make this logic as a part of the `HermesReactor` so it will be possible to maintain this execution loop of the cron events ?
Same as it was done for the hermes event queue

hermes/bin/src/runtime_extensions/hermes/cron/event.rs
saibatizoku: CronTagged is defined in a WIT file within the project, I don't think that it's possible to use the `derive` macro.

hermes/bin/src/runtime_extensions/hermes/cron/state.rs
saibatizoku: Actually... thinking a bit more about this, I think that we should keep the join handle and make it more maintainable, but lets talk about how coupled the HermesReactor should be to the module APIs to discuss the best approach.

Mr-Leshiy (approved):
LGTM

hermes/bin/src/runtime_extensions/hermes/cron/event.rs
```diff
@@ -1,15 +1,23 @@
...
+use chrono::Utc;
+use saffron::Cron;
+
+use super::state::cron_queue_rm;
 use crate::{
     event::HermesEventPayload, runtime_extensions::bindings::hermes::cron::api::CronTagged,
 };
 
+/// Timestamp for when to run the cron.
+pub(crate) type CronTimestamp = u64;
```
minikin, line 12: Can we consider using a _newtype pattern_ here instead of _type alias_?

minikin (changes requested):

minikin (approved):
LGTM

PR #157, opened 2024-02-23
feat: follower connection

cong-or:
Currently if the follower cannot find the mithril snapshot, it just returns an error `Error: MithrilSnapshot `

However, I was hoping we could abstract this failure and return a follower regardless i.e if it cant bootstrap from the snapshot it should return a follower which is consuming from a network. 

Maybe i'm wrong and there's an easier cleaner way, but I had to do the following to mimic the behavior described above on the consumer side.

The PR requires the `FollowerConfig` is **cloneable** for this code to work because of the move.

We can just use the below code for now (if there is no other way) and add the suggestions to the follower at a later date or address it soon on the follower side of things.

I could be way off and missing something. Please let me know.

```rust 

/// In the context of setting up the follower connection.
/// If there is metadata present which allows us to bootstrap from a point in time
/// We start from there, if not; we start from genesis.
async fn follower_connection(
    start_from: (Option<SlotNumber>, Option<BlockHash>), snapshot: &str, network: Network,
    relay: &str,
) -> Result<Follower, Box<dyn Error>> {
    let mut follower_cfg = if start_from.0.is_none() || start_from.1.is_none() {
        // start from genesis, no previous followers, hence no starting points.
        FollowerConfigBuilder::default()
            .mithril_snapshot_path(PathBuf::from(snapshot))
            .build()
    } else {
        // start from given point
        FollowerConfigBuilder::default()
            .follow_from(Point::new(
                start_from.0.ok_or("Slot number not present")?.try_into()?,
                hex::decode(start_from.1.ok_or("Block Hash not present")?)?,
            ))
            .mithril_snapshot_path(PathBuf::from(snapshot))
            .build()
    };

    let follower = match Follower::connect(&relay, network, follower_cfg).await {
        Ok(follower) => follower,
        Err(err) => {
            error!(
                "Unable to bootstrap via mithril snapshot {}. Trying network..",
                err
            );
           
            // We know bootstrapping from the snapshot fails, remove path for it to try from network
            follower_cfg.mithril_snapshot_path = None;
            Follower::connect(&relay, network, follower_cfg).await?
        },
    };

    Ok(follower)
}



```

### DIFF
--- a/hermes/crates/cardano-chain-follower/examples/follow_chain_updates_mithril.rs
+++ b/hermes/crates/cardano-chain-follower/examples/follow_chain_updates_mithril.rs
@@ -41,7 +41,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     .await?;
 
     // Wait for some chain updates and shutdown.
-    loop {
+    for _ in 0..10 {
         let chain_update = follower.next().await?;
 
         match chain_update {
@@ -67,6 +67,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
             },
         }
     }
+
+    follower.close().await?;
+
+    Ok(())
 
     // Waits for the follower background task to exit.
 }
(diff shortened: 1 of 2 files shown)


FelipeRosa (approved):
LGTM

PR #174, opened 2024-03-07
docs(hermes): Document how the follower runtime extension manages instances

### LINKED ISSUE #177: [Bug]: CI broken due to flaky kroki.io server
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

We are getting 5xx errors on some files when trying to build docs with images rendered using the kroki.io service.

### Expected Behavior

Docs don't fail because of kroki.io

### Project

Documentation

### What operating system are you seeing the problem on?

Other

### What browsers are you seeing the problem on?

Not Applicable

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stevenj:
# Description

Documentation on how instances of the Hermes node follower are managed by the runtime extension.

Closes #177


docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
```diff
@@ -0,0 +1,122 @@
...
+When a wasm module in an application subscribes to a Node followers data stream, IF there is a pre-existing node follower
+which can serve its needs, then it should simply receive the events from that stream.
+
+It will not create a new stream for every subscription.
+
+When the last subscriber to a Node followers connection/stream unsubscribes, it will stop the connection.
+
+This means that there is not one follower per application, there is one follower per stream type.
+
+For example,  if a wasm module subscribes to a stream of events from Genesis, and Tip, it will have two active streams.
```
FelipeRosa, line 26: A tip follower is a special case in the sense that one instance can be shared by all subscriber modules that want to subscribe to tip events.

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
```diff
@@ -0,0 +1,122 @@
...
+For the initial implementation of the Follower, we will employ technique 2 and a limited version of Technique 3 ONLY.
+
+#### Delaying starting of the streams
+
+When a stream is subscribed, and it results in a new stream, then the connection to that relay node will not happen for 5 seconds.
+Any further subscriptions to the same stream can then be added to the pending connection.
+Once the connection is made, then further subscriptions will only be added to it if they are valid for that connection.
+Otherwise a second connection is established, again with its own 5 second start delay.
+
+This delay occurs regardless of where the stream is to start from Tip, Genesis or a particular block.
```
FelipeRosa, line 80: What if instead of just waiting 5 seconds we kept track of where each follower is at and on a subscription request we:

1. assigned the closest one that is at a point previous to the point the subscriber wants to follow from
2. Waited for 5 seconds for it to reach the point the subscriber wants
3. If it didn't reach that point in 5 seconds, we spawn another follower at that point

Does that make sense?

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
```diff
@@ -0,0 +1,122 @@
...
+The purpose of this is to ensure that once followers have fully synced we are not running excessive followers
+that are doing the same work and reporting the same events.
+
+### Handling "Continue" connections
+
+There is an option to listen to a stream from "Continue" but what does "Continue" actually mean?
+If there are multiple streams it could match on, which one does it match?
+
+We will use the rule that if "Continue" is specified, it means "Follow the blockchain with the earliest blocks".
+
```
FelipeRosa, line 100: Storing the point each follower is at would be used at this point as well as my comment above. I guess it would also allow the runtime extension to find a follower for a module on a "continue" request.

We could also make it reuse the logic I mentioned above for not only finding the follower with the earliest blocks but to wait and, possibly, spawn a follower for conitnuing at the point it was stopped. But this is just a thought.

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
```diff
@@ -0,0 +1,122 @@
...
+blockchain followers running or pending for the same Application?
+This can only happen either at the start of world for a particular Application OR
+an Application has unsubscribed from all blockchain events.
+
+In this case the "Continue" subscription should be added to list of continue subscriptions for each application.
+When a block is received and about to be sent as an event to an application, we can check the "Continue" list for that app,
+
+If there are subscriptions in that list,
+and the event is from the oldest active follower for the app then they are removed from that list,
+and added as active subscriptions to the selected follower.
```
FelipeRosa, line 114: I didn't get this because if none of the app's modules are subscribed yet and one of them wants to "continue", should the runtime extension find (or spawn) a follower to follow from either the genesis (if it's an unitialized app) or the point where it previously stopped/unsubscribed?

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
stevenj: Yes,  but its also just a special case of multiple subscriber modules wanting to subscribe to the same point in the stream.
It's just that this is going to be highly likely so we should accommodate it now.
But the general case is anytime two followers end up on the same block on the same network, they can be collapsed into a single follower.

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
stevenj: That only works if your asfter somethin that's in the future.  If I request Genesis, and the follower immediately gets block 1 and 2 and sends those events, its now a follower pointing to block 3.
If another module asks to follow from genesis our existing follower is already past genesis so we now need another follower that's running 3 blocks behind.  The delay is to ensure we don't get that scenario, which is more likely.

For example, I have a module that says "Give me the blocks from genesis", and then a different module says "Give me the transactions from genesis".  We could end up with two followers, when what we really wanted was one.

stevenj: there are other ways, but a 5 second delay is an easy one to ensure followers requested at about the same time are synced.  Your suggestion would also be viable if I asked for genesis in one follower, and block 10 in another, we could park the block 10 one until the genesis follower catches up.  But I think this is a less likely scenario, though worth documenting.

What I am not trying to achieve here is optimal follower setup, but one that catches the most likely events and then we can see how it behaves and iterate.  I don't want to over think/over engineer it too much up front.

In fact in a first PR I would 100% accept no optimization of the followers at all, just so we get the logic right in its simplest form.  Then adding progressive optimizations to the running followers could be follow up.  Also for M2 we don;t need these optimizations.

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
stevenj: Yes each follower will need to know where its at for "continue" to work in its simplest case.
See above for not needing to do these optimizations yet, except for probably tracking where a follower is at so we can properly implement "continue" in a simple way initially.

docs/src/architecture/08_concepts/cardano_chain_follower/follower_instances.md
stevenj: We don't know what order modules will start in,  so its possible that one module is designed to follow the stream of another.  I am thinking that "follow" might be not sufficient after thinking this through.

For now I think we should do the following, because I am not 100% sure `continue` as its currently cast is workable.
Just comment out `continue`, we never said it would be an option for M2.  That simplifies things a lot.
We can then get the follower working in a naive way very easily, we just need to track what modules are following what, and we can have a unique follower for each possibility in our first iteration.
Clearly its not optimal, but it would work ok.

stevenj: This document can stand as is, because its a design document.  We don't need to implement all of it in your current work, and we can revise it as needed.

saibatizoku (approved):

PR #181, opened 2024-03-12
Implement Cardano chain follower functions in Cardano Runtime Extension

FelipeRosa:


hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+/// Reads a block from a Cardano network.
+pub(super) fn read_block(
+    chain_id: CardanoBlockchainId, at: cardano_chain_follower::PointOrTip,
+) -> Result<cardano_chain_follower::MultiEraBlockData> {
+    STATE.tokio_rt_handle.read_block(chain_id, at)
+}
+
+/// Runs the Cardano Runtime Extension Tokio runtime.
+#[instrument(skip(cmd_rx))]
+fn tokio_runtime_executor(mut cmd_rx: TokioRuntimeHandleCommandReceiver) {
```
minikin, line 345: Can we break down this function into smaller parts?

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+                },
+            }
+        }
+    });
+}
+
+/// Runs a Cardano chain follower that generates events for the given application module
+/// and is connected to the given chain.
+#[instrument(skip(cmd_rx, follower), fields(app_name = %app_name, module_id = %module_id))]
+async fn chain_follower_executor(
```
minikin, line 459: Can we decompose this function into smaller functions?

minikin (changes requested):

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+        /// Chain point at which the block is to be fetched.
+        at: cardano_chain_follower::PointOrTip,
+        /// Response channel sender.
+        response_tx:
+            tokio::sync::oneshot::Sender<Result<cardano_chain_follower::MultiEraBlockData>>,
+    },
+}
+
+/// Tokio runtime handle command channel sender type.
+type TokioRuntimeHandleCommandSender = tokio::sync::mpsc::Sender<TokioRuntimeCommand>;
```
Mr-Leshiy, line 57:
```suggestion
type TokioRuntimeCommandSender = tokio::sync::mpsc::Sender<TokioRuntimeCommand>;
```

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+
+/// Tokio runtime handle command channel receiver type.
+type TokioRuntimeHandleCommandReceiver = tokio::sync::mpsc::Receiver<TokioRuntimeCommand>;
```
Mr-Leshiy, line 60:
```suggestion
type TokioRuntimeCommandReceiver = tokio::sync::mpsc::Receiver<TokioRuntimeCommand>;
```

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+        >,
+    ),
+    /// Instructs the chain follower to stop generating events.
+    Stop(tokio::sync::oneshot::Sender<()>),
+    /// Instructs the chain follower to resume generating events.
+    Continue(tokio::sync::oneshot::Sender<()>),
+}
+
+/// Chain follower handle command channel sender.
+type ChainFollowerHandleCommandSender = tokio::sync::mpsc::Sender<ChainFollowerCommand>;
```
Mr-Leshiy, line 139:
```suggestion
type ChainFollowerCommandSender = tokio::sync::mpsc::Sender<ChainFollowerCommand>;
```

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+/// Chain follower handle command channel receiver.
+type ChainFollowerHandleCommandReceiver = tokio::sync::mpsc::Receiver<ChainFollowerCommand>;
```
Mr-Leshiy, line 141:
```suggestion
type ChainFollowerCommandReceiver = tokio::sync::mpsc::Receiver<ChainFollowerCommand>;
```

Mr-Leshiy (commented):
Just a general question to be on the same page 😊.
What is the difference between TokioRuntime and ChainFollowerRuntime ?
Why do not combine them under the same abstractions ?

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
```diff
@@ -1,7 +1,788 @@
...
+    tokio_rt_handle: TokioRuntimeHandle,
+    /// Mapping of application module subscription states.
+    subscriptions:
+        DashMap<(HermesAppName, ModuleId, cardano_chain_follower::Network), SubscriptionState>,
+    /// Chain followers configured only for reading blocks.
+    readers: DashMap<cardano_chain_follower::Network, cardano_chain_follower::Follower>,
+}
+
+/// Cardano Runtime Extension internal state.
+static STATE: once_cell::sync::Lazy<State> = once_cell::sync::Lazy::new(|| {
```
Mr-Leshiy, line 227: What do you think to move initialisation as a part of the `HermesReactor` and place a thread `JoinHandle` as a field there, so it reactor will maintain an execution loop of the follower.
The same as it was done for hermes event queue

FelipeRosa:
> Just a general question to be on the same page 😊. What is the difference between TokioRuntime and ChainFollowerRuntime ? Why do not combine them under the same abstractions ?

@Mr-Leshiy 

There's 1 OS thread for running a Tokio runtime which is needed because the `cardano-chain-follower` crate uses `async` code. I'm using a channel here in the `TokioRuntimeHandle` to send commands to the OS thread running the Tokio runtime.

Then there is the same pattern for communicating with each running follower. For now, there's 1 handle per follower since we decided to not do any optimizations yet.

I don't think it makes sense to merge them into the same abstraction since they serve different purposes in the Cardano Runtime Extension.

hermes/bin/src/runtime_extensions/hermes/cardano/mod.rs
FelipeRosa: I think holding the thread join handle somewhere is a good ideia. But I think it would be better that we had some function that the reactor would call when it's time for the runtime extension to be shutdown instead of leaking the logic of runtime extensions on the reactor. 

I mean, having the logic for the queue in the reactor makes sense because the event queue is managed by the reactor but I don't think the reactor should have code for every background thread we may want to spawn.

WDYT?

saibatizoku: I'm thinking along the same lines for the cron queue.

FelipeRosa: Yeah, I agree that we should gracefully shutdown Hermes. But I think the reactor should signal the Runtime Extensions that they should shutdown. Just waiting on a `JoinHandle` is not enough for that.

In the case the background thread is running a loop (like the Cardano Runtime Extension) there needs to be some kind of communication (or something like dropping channel receivers/senders) in order for the loop to be stopped. Otherwise the reactor would wait forever.

hermes/bin/src/runtime_extensions/hermes/cardano/chain_follower_task.rs
```diff
@@ -0,0 +1,295 @@
...
+        cmd_rx, follower, app_name, module_id, chain_id,
+    ));
+
+    Handle { cmd_tx }
+}
+
+/// Runs a Cardano chain follower that generates events for the given application module
+/// and is connected to the given chain.
+#[instrument(skip(cmd_rx, follower), fields(app_name = %app_name, module_id = %module_id))]
+async fn executor(
```
minikin, line 108: What do you think about refactoring this function to something like this?

```rust
#[instrument(skip(cmd_rx, follower), fields(app_name = %app_name, module_id = %module_id))]
async fn executor(
    mut cmd_rx: CommandReceiver, mut follower: cardano_chain_follower::Follower,
    app_name: HermesAppName, module_id: ModuleId, chain_id: CardanoBlockchainId,
) {
    let network = chain_id.into();
    let module_state_key = (app_name, module_id, network);

    let mut stopped = false;

    'exec_loop: loop {
        tokio::select! {
            res = cmd_rx.recv() => {
                let Some(cmd) = res else {
                    break 'exec_loop;
                };

                match cmd {
                    Command::SetReadPointer(follow_from, res_tx) => {
                        stopped = set_read_pointer(&mut follower, follow_from).await;
                        drop(res_tx.send(stopped));
                    }
                    Command::Stop(res_tx) => {
                        stopped = true;
                        let _ = res_tx.send(());
                    }
                    Command::Continue(res_tx) => {
                        stopped = false;
                        let _ = res_tx.send(());
                    }
                }
            }

            result = follower.next(), if !stopped => {
                stopped = process_chain_update(&module_state_key, chain_id, result).await;
            }
        }
    }

    drop(follower.close().await);
}

async fn set_read_pointer(
    follower: &mut cardano_chain_follower::Follower, follow_from: FollowFrom,
) -> bool {
    let result = follower.set_read_pointer(follow_from).await;

    match &result {
        Ok(Some(point)) => {
            trace!(slot = point.slot_or_default(), "Follower read pointer set");
            false
        },
        Ok(None) => {
            warn!("Couldn't set follower read pointer: point not found");
            true
        },
        Err(e) => {
            warn!(error = ?e, "Failed to set read pointer");
            true
        },
    }
}

async fn process_chain_update(
    module_state_key: &ModuleStateKey, chain_id: CardanoBlockchainId,
    chain_update: Result<cardano_chain_follower::ChainUpdate>,
) -> bool {
    match chain_update {
        Ok(chain_update) => {
            match chain_update {
                cardano_chain_follower::ChainUpdate::Block(block_data) => {
                    process_block_update(module_state_key, chain_id, block_data).await;
                },
                cardano_chain_follower::ChainUpdate::Rollback(block_data) => {
                    process_rollback_update(module_state_key, block_data).await;
                },
            }
            false
        },
        Err(e) => {
            warn!(error = ?e, "Failed to get chain update");
            true
        },
    }
}

async fn process_block_update(
    module_state_key: &ModuleStateKey, chain_id: CardanoBlockchainId,
    block_data: cardano_chain_follower::BlockData,
) {
    let (subscribed_to_blocks, subscribed_to_txns, _) = get_subscriptions(module_state_key);

    if subscribed_to_blocks {
        generate_block_event(module_state_key, chain_id, block_data).await;
    }

    if subscribed_to_txns {
        generate_txn_events(module_state_key, chain_id, block_data).await;
    }
}

async fn process_rollback_update(
    module_state_key: &ModuleStateKey, block_data: cardano_chain_follower::BlockData,
) {
    let (_, _, subscribed_to_rollbacks) = get_subscriptions(module_state_key);

    if subscribed_to_rollbacks {
        generate_rollback_event(module_state_key, block_data).await;
    }
}

fn get_subscriptions(module_state_key: &ModuleStateKey) -> (bool, bool, bool) {
    if let Some(sub_state) = STATE.subscriptions.get(module_state_key) {
        (
            sub_state.subscribed_to_blocks,
            sub_state.subscribed_to_txns,
            sub_state.subscribed_to_rollbacks,
        )
    } else {
        (false, false, false)
    }
}

async fn generate_block_event(
    module_state_key: &ModuleStateKey, chain_id: CardanoBlockchainId,
    block_data: cardano_chain_follower::BlockData,
) {
    let on_block_event = super::event::OnCardanoBlockEvent {
        blockchain: chain_id,
        block: block_data.into_raw_data(),
        source: BlockSrc::NODE,
    };

    if let Err(err) = crate::event::queue::send(HermesEvent::new(
        on_block_event,
        TargetApp::List(vec![module_state_key.0.clone()]),
        TargetModule::_List(vec![module_state_key.1.clone()]),
    )) {
        error!(error = ?err, "Failed to send Cardano block event to the Event queue");
    } else {
        trace!(
            block_number = block_data.decode().map(|b| b.number()).unwrap_or_default(),
            "Generated Cardano block event"
        );
    }
}

async fn generate_txn_events(
    module_state_key: &ModuleStateKey, chain_id: CardanoBlockchainId,
    block_data: cardano_chain_follower::BlockData,
) {
    let txs = block_data.decode().map(|b| b.txs()).unwrap_or_default();

    for (tx, index) in txs.iter().zip(0u32..) {
        let on_txn_event = super::event::OnCardanoTxnEvent {
            blockchain: chain_id,
            slot: block_data.decode().map(|b| b.slot()).unwrap_or_default(),
            txn_index: index,
            txn: tx.encode(),
        };

        if let Err(err) = crate::event::queue::send(HermesEvent::new(
            on_txn_event,
            TargetApp::List(vec![module_state_key.0.clone()]),
            TargetModule::_List(vec![module_state_key.1.clone()]),
        )) {
            error!(error = ?err, "Failed to send Cardano transaction event to the Event queue");
        } else {
            trace!(
                block_number = block_data.decode().map(|b| b.number()).unwrap_or_default(),
                tx_count = txs.len(),
                "Generated Cardano block transactions events"
            );
        }
    }
}

async fn generate_rollback_event(
    module_state_key: &ModuleStateKey, block_data: cardano_chain_follower::BlockData,
) {
    let on_rollback_event = super::event::OnCardanoRollback {
        blockchain: CardanoBlockchainId::Preprod,
        slot: block_data.decode().map(|b| b.slot()).unwrap_or_default(),
    };

    if let Err(err) = crate::event::queue::send(HermesEvent::new(
        on_rollback_event,
        TargetApp::List(vec![module_state_key.0.clone()]),
        TargetModule::_List(vec![module_state_key.1.clone()]),
    )) {
        error!(error = ?err, "Failed to send Cardano block event to the Event queue");
    } else {
        trace!(
            block_number = block_data.decode().map(|b| b.number()).unwrap_or_default(),
            "Generated Cardano rollback event"
        );
    }
}

```

hermes/bin/src/runtime_extensions/hermes/cardano/chain_follower_task.rs
```diff
@@ -0,0 +1,295 @@
...
+    mut cmd_rx: CommandReceiver, mut follower: cardano_chain_follower::Follower,
+    app_name: HermesAppName, module_id: ModuleId, chain_id: CardanoBlockchainId,
+) {
```
minikin, line 111: Can we consider replacing ` 'exec_loop: loop` with something like `while !stopped`(_see the draft below_)?

* **Readability and Intent** 
Using `while !stopped` immediately conveys the loop's continuation condition directly in the loop's declaration, making it clear to readers that the loop will continue executing as long as `stopped` is `false`. This directly ties the loop's execution to the state of a specific variable (stopped), making the code easier to understand.

* **Maintains Focus on Business Logic**
This can simplify the loop control structure and allow future maintainers or readers to focus more on the business
 logic within the loop rather than the mechanics of loop control. 
 
*  **Improves Code Clarity**
The `while !stopped` loop makes the control logic dependent on the state of the stopped variable very explicit. It eliminates the need for readers to identify where and how the loop might be terminated. This direct association between the loop continuation and the stopped variable enhances clarity by immediately making the loop's termination condition visible.

```rust
#[instrument(skip(cmd_rx, follower), fields(app_name = %app_name, module_id = %module_id))]
async fn executor(
    mut cmd_rx: CommandReceiver, mut follower: cardano_chain_follower::Follower,
    app_name: HermesAppName, module_id: ModuleId, chain_id: CardanoBlockchainId,
) {
    let network = chain_id.into();
    let module_state_key = (app_name, module_id, network);

    let mut stopped = false;

    while !stopped {
        tokio::select! {
            res = cmd_rx.recv() => {
                // Improved pattern matching for clarity.
                match res {
                    Some(cmd) => {
                        match cmd {
                            Command::SetReadPointer(follow_from, res_tx) => {
                                // Handling the potential error from set_read_pointer asynchronously.
                                if let Err(e) = set_read_pointer(&mut follower, follow_from).await {
                                    eprintln!("Error setting read pointer: {:?}", e);
                                    // Consider if stopping on error is desired.
                                }
                                // Now explicitly ignoring the send result but consider logging the error.
                                let _ = res_tx.send(stopped);
                            }
                            Command::Stop(res_tx) => {
                                stopped = true;
                                let _ = res_tx.send(());
                            }
                            Command::Continue(res_tx) => {
                                stopped = false;
                                let _ = res_tx.send(());
                            }
                        }
                    },
                    None => break, // Cleaner break from the loop if channel is closed.
                }
            }

            result = follower.next(), if !stopped => {
                // Error handling for process_chain_update is similar to set_read_pointer.
                if let Err(e) = process_chain_update(&module_state_key, chain_id, result).await {
                    eprintln!("Error processing chain update: {:?}", e);
                    // Consider if stopping on error is desired.
                }
            }
        }
    }

    // Improved resource cleanup with error handling.
    if let Err(e) = follower.close().await {
        eprintln!("Error closing the follower: {:?}", e);
    }
}

```

minikin (changes requested):

hermes/bin/src/runtime_extensions/hermes/cardano/chain_follower_task.rs
FelipeRosa: I this case we don't want to stop the loop if `stopped = false`. 

`stopped` will be `false` when the app module that is using this follower wants to stop the Cardano Runtime Extension event generation. In that case, we still want to wait for commands without progressing the following process.

The loop will only be broken if there are no command `Sender`s or the module's subscription state is removed from the extension's state.

I used a label to name the loop and make it clear when using `break`.

minikin: Thank you, @FelipeRosa, for the clarifications.

minikin (approved):
```
██╗      ██████╗ ████████╗███╗   ███╗
██║     ██╔════╝ ╚══██╔══╝████╗ ████║
██║     ██║  ███╗   ██║   ██╔████╔██║
██║     ██║   ██║   ██║   ██║╚██╔╝██║
███████╗╚██████╔╝   ██║   ██║ ╚═╝ ██║
╚══════╝ ╚═════╝    ╚═╝   ╚═╝     ╚═╝
```

PR #184, opened 2024-03-19
fix(cardano-chain-follower): set_read_pointer now returns None when the requested point is not found

FelipeRosa:
# Description

This PR fixes a bug I found while testing the Cardano Runtime Extension.

Calling `Follower::set_read_pointer` should return `Some(point)` if the requested point was found on chain (and the follower is now following from that point) and `None` if the point was not found. The current implementation always returns `None`.


cong-or (approved):
lgtm

PR #198, opened 2024-04-04
feat(warm): Add Cardano Runtime Extension integration test skeleton

FelipeRosa:
This PR adds the integration tests skeleton for the Cardano Runtime Extension. 

The current implementation only tests fetching blocks from the network through the Cardano Runtime Extension's API. It fetches a block, parses it using `pallas-traverse` and does some assertions.


stevenj (approved):
Very nice work.
Good job.

PR #206, opened 2024-04-09
feat: Mithril Snapshot Baseline Stats

### LINKED ISSUE #201: 🛠️ [TASK] : Baseline mithril performance integration tests
### Summary

Produce baseline integration tests to measure optimal mithril sync perf

### Description

When syncing from mithril we need a way to determine relative performance from an optimal baseline.
We also need a way to measure the overhead to that our follower crate adds.
So we need two integration tests for the follower crate.
Both tests are to be disabled by default in an integration test run, and only run when requested.
We need two earthly tagets which set up and run these two integration tests (again not as default tests, so don't use a CI default target name).

The tests require the mithril snapshot from:
https://github.com/input-output-hk/catalyst-ci/blob/9ada7bce41dedbc8cbe2b806591ac4e6e6b1cb08/earthly/mithril_snapshot/Earthfile#L21

They run until it has been exhausted.

In order to use these tests as meaningful analysis tools, we need a set of consistent stats.
They need to produce output statistics similar to this as they run (20 second interval on stats, fails if the end of the mithril snapshot has not been reached and no more slots are processed for 60 seconds):

```
./c/t/api_tests+test | 2024-04-07 04:00:07.799 | INFO     | api_tests.test_utxo:sync_to:25 -  synced to slot  :  in total time   : slots/sec (interval) : slots remaining  : est. time to go  :
./c/t/api_tests+test | 2024-04-07 04:00:07.800 | INFO     | api_tests.test_utxo:sync_to:63 -            91080 :         00:00:00 :     818.08(91082.00) :         56273094 :         19:06:26 :
./c/t/api_tests+test | 2024-04-07 04:00:27.802 | INFO     | api_tests.test_utxo:sync_to:63 -           257000 :         00:00:20 :     8294.78(8295.28) :         56107174 :         01:52:44 :
```

The two integration tests are essentially identical, the only difference is where the data is sourced.

1. Using Pallas directly, read every block from the snapshot as fast as possible.  Produce the stats as shown above on 20s intervals, and a summary at the end, or a failure if it stops.
2. Using the API of the chain follower, do the same.

So the only difference is if we are using pallas directly, or the chain follower API.

The first analysis we will perform here is if there is a performance delta between raw pallas and the chain follower.

These tests and alanalysis will feed into tests being don in catalyst-gateway and are critical to help understand issues we are seeing there.

saibatizoku:
# Description

Produce baseline integration tests to measure optimal mithril sync performance.

## Related Issue(s)

Closes #201 .

## Description of Changes


stevenj (changes requested):
Can i get you to target the `feat/m3` branch rather than `main` for this PR?

hermes/crates/cardano-chain-follower/examples/load_snapshot.rs
```diff
@@ -0,0 +1,60 @@
...
+            tracing::info!(
+                slots_seen = slots_seen,
+                total_time = total_duration,
+                "NUMBER={} SLOT={} HASH={}",
+                block.number(),
+                block.slot(),
+                hex::encode(block.hash())
+            );
+            timer = Instant::now();
+        }
```
stevenj, line 51: The stats seem off.  `slots_seen=2118401 total_slots_expected=56722224` looks like we missed data/
is it that `slots_seen` should be `blocks_seen`?
Can we also make the output columnar, and add summary data to the line, such as slots/second, blocks/second (aggregate since the start, and also just for the current interval.)

For example more like this:
```
BLOCK   |   SLOT    |   Total Time |  Blocks/sec | blocks/sec/int | slots/sec | slots/sec/int |
792223 |  24833596 | 00:00:10 | 79222.3 | 79222.3 | 2483359.6 | 2483359.6 | 
1318069 | 37574044 | 00:00:20 | 65,903.45 | 52584.6 | 1878702.2 | 1274044.8 | 
```
And so on, but pad the columns so they are better aligned.
`blocks/sec/int` means blocks/second in the last interval or since the last time we printed.  The other stat is cumulative since the start of the run.

That data in the example is from my run of this test.  I want the summarily stats so I don;t have to guess how its performing over time or manually calculate them.

It does looks like its actually getting slower over time.  

To determine if that's the case, also add up the total size of all the blocks, and include a 
`block size` metric, and a `block MB/sec` and `block MB/sec/int` summary metric to the line.
It could be early blocks are tiny, so we can do lots but later they are big, so that's more data to process.
If the number of blocks/slots changes but the MB/s metrics don't we know that's the cause.

The data set is also not big enough to determine that.
Lets add a "prod" snapshot target to cat-ci which we can also use to run this test against.
The URL you can use for it is: https://storage.googleapis.com/cdn.aggregator.release-mainnet.api.mithril.network/mainnet-e477-i5604.d11e9b0a3cda3f32233cf7c89f0a1aa321182c790b3277af468c8051007f89b7.tar.zst
Note: that's 48GB so for now we would only use that in local testing, not in CI.  But we can still have the target in earthly.

You can different targets in the earthfile which use a different snapshot.

As you have made this a stand alone app, which I am not opposed to, I think we can just extend this to have a simple option (no need for complex cli, just read the first parameter) to make it run `pallas` or `hermes` mode.

`pallas` does what you do here.
`hermes` mode uses the chain follower crate to do exactly the same thing.

The metrics can be the same then, the only difference is how you source the blocks.

stevenj (changes requested):

hermes/crates/cardano-chain-follower/examples/load_snapshot.rs
saibatizoku: Sorry, I hadn't uploaded the most recent code, I'll take a look at your comments and address them.

minikin:
@saibatizoku, please reopen PR against main.

saibatizoku:
Superseded by #211 .

PR #211, opened 2024-04-17
feat(cardano-chain-follower): Overhead benchmark implementations (pallas + cardano-chain-follower)

### LINKED ISSUE #201: 🛠️ [TASK] : Baseline mithril performance integration tests

FelipeRosa:
### Description

*NOTE: 
This replaces #206. I have added @saibatizoku as a co-author in the commits as his draft PR served as inspiration for this.*

This PR implements a benchmark to measure the overhead that the internal implementation of the `cardano-chain-follower` has in comparison to using the pallas library directly to read a Mithril snapshot.

### Related Issues

Closes #201.


minikin (approved):
LGTM

@FelipeRosa  We need to consider publishing benchmark results in the future.
What do you think?

FelipeRosa:
> LGTM
> 
> @FelipeRosa We need to consider publishing benchmark results in the future. What do you think?

I agree. We need to define how we are going to do that because running the benchmarks also involves downloading/storing Mithril snapshots which, for mainnet specially, can be big. Maybe we want to consider triggering them manually or nightly (something like that).

minikin:
task to publish results  #217

PR #236, opened 2024-05-20
feat(p2p): Hermes Peer2Peer Communications with IPFS/LibP2P

saibatizoku:
# Description

Add stand-alone `hermes-ipfs` crate to provide functionality related to p2p and IPFS.

## Related Issue(s)

* #242 
* #243 

## Description of Changes

Provides API for pin/unpin, add/get file, p2p pubsub/kv networking, and a validation mechanism for data.

## Related Pull Requests

* #245 


hermes/crates/hermes-ipfs/examples/provide-content-dht.rs
```diff
@@ -0,0 +1,113 @@
+//! Use Hermes IPFS to distribute content using DHT
```
FelipeRosa, line 1: I think this should be using `dht_put` and `dht_get`? https://docs.rs/rust-ipfs/0.11.20/rust_ipfs/struct.Ipfs.html#method.dht_put

saibatizoku: I'll add another example with these methods, also to show how to use the underlying `ipfs` node.

hermes/crates/hermes-ipfs/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+/// Hermes IPFS
+#[allow(dead_code)]
+pub struct HermesIpfs {
+    /// IPFS node
+    pub node: Node,
```
Mr-Leshiy, line 54: What is the purpose to make this field public ? Seems like it should a private.
Also if change public to private for this field, seems like make sense also get rid of the `Node` struct and use just a plain `Ipfs` type here.

hermes/crates/hermes-ipfs/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+    ///
+    /// ## Returns
+    ///
+    /// * `A result with Vec<u8>`.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the file fails to download.
+    pub async fn get_ipfs_file(&self, ipfs_path: GetIpfsFile) -> anyhow::Result<Vec<u8>> {
+        let stream_bytes = self.node.cat_unixfs(ipfs_path).await?;
```
Mr-Leshiy, line 113: Is it possible to return here just a bytes stream ?
For the case when a file is really huge I think this approach could lead to performance issues.

hermes/crates/hermes-ipfs/src/lib.rs
saibatizoku: The purpose is to have access to low-level functionality that the inner node provides. This is useful for more complex use cases, as can be seen in `examples/pubsub.rs`. I think it's preferrable to have our new type for this, and not re-use Ipfs directly, as we might want to swap the implementation in the future.

hermes/crates/hermes-ipfs/src/lib.rs
saibatizoku: This method should just return the resulting bytes from IPFS, under the hood, it is implemented as an async stream. However, if the use case for handling the stream is required, the user can invoke `self.node.cat_unixfs` directly, and deal with it as needed.

hermes/crates/hermes-ipfs/src/lib.rs
Mr-Leshiy: I see your point.
But from my point of view we are trying to give for the user too much flexibility and responsibility.
Actually deferring the choice how it should be used.
I think we should hide such internal details, just to be safe. Also current interface looks exhaustive already and even if we will need something we can easily extend it.
But I dont have a really strong position here, I just feel that it should be private 😊

saibatizoku: I can't think of any reason to defend it being public. I've updated the code.

hermes/crates/hermes-ipfs/src/lib.rs
saibatizoku: I think this should stay as it is, and if we find issues we can revise.

Mr-Leshiy (approved):
LGTM

stevenj (approved):
LGTM

Issue #237, opened 2024-05-21
🛠️ [TASK] : Wasm module packaging


Mr-Leshiy:
### Summary

Implement a packaging functionality of wasm module package following by this spec https://input-output-hk.github.io/hermes/architecture/08_concepts/hermes_packaging_requirements/wasm_modules/#overview-of-a-wasm-component-module

### Description

Following the spec, need the following functionality:
- [x] By the provided  directory path, build an hdf5 package for wasm module with all needed data init.
- [x] Apply compression to all data inside hdf5 package.
- [ ] Validate all the contents during packaging and loading process.

PR #238, opened 2024-05-21
feat: Add basic wasm module packaging

Mr-Leshiy:
# Description

Added a basic implementation of creating a wasm module hdf5 package from the provided directory.
It is a basic implementation, without any validation, just coping appropriate files to the hdf5 package.

## Related Issue(s)

#237

## Description of Changes

- Implemented basic module wasm hdf5 package building from the provided directory
- Updated `cli::Package` command with the added implementation
- Added `Errors` struct which is a new error handling abstraction which helps to collect many errors and return them all at the end of the function execution. It solved the problem of early return and collect all possible issues and return them all.
- Added fancy printing during the cli execution.


hermes/bin/src/cli/package.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[clap(long)]
+    dir: PathBuf,
+}
+
+impl PackageCommand {
+    /// Run cli command
+    pub(crate) fn exec(self) -> anyhow::Result<()> {
+        println!("{} Build wasm module package...", Emoji::new("📦", ""));
+
+        WasmModulePackage::from_dir(self.dir)?;
```
stevenj, line 24: Just throwing errors out the top of the application isn't acceptable.
It is not a good practice, and we need to properly handle them and report them.  Not just rely on a default "Print the error struct" catcher.

currenty this is doing this:
```
target/release/hermes package --dir /dosnt/exist
ℹ️
version: 0.0.1,
git info: {49e67f7308cf8d26d7d48b3fbb03423b12c10442.+ (feat/wasm-packaging)}
compiler: rustc 1.78.0 (9b00956e5 2024-04-29)
build time: 2024-05-22 04:06:02Z

📦 Build wasm module package...
🚨:
- H5Fcreate(): unable to create file: unable to open file: name = '/dosnt/exist/module.hdf5', errno = 2, error message = 'No such file or directory', flags = 13, o_flags = 242
```

We should be testing for the appropriate errors that can happen in normal use (like pointing to a directory that doesn't exist), and give a reasonable error message that explains why we failed and what they should do about it.

The error it currently gives is hostile and only a programmer would have any hope of understanding what it means.

stevenj: To be clear, "unexpected" errors can print a message like this, but they should still be caught and displayed in a nicer way than this.  This should only be for "unexpected" errors though, usually generated by a bug.  Errors that are possible due to bad parameters, or files are to be properly caught and handled.

hermes/bin/src/packaging/wasm_module.rs
```diff
@@ -0,0 +1,115 @@
...
+    const SETTINGS_SCHEMA_JSON: &'static str = "settings.schema.json";
+    /// Share directory name.
+    const SHARE: &'static str = "share";
+
+    /// Create a new Wasm module package from a directory.
+    pub(crate) fn from_dir<P: AsRef<Path>>(path: P) -> anyhow::Result<Self> {
+        let path = path.as_ref();
+        let mut errors = Errors::new();
+
+        let package_path = path.join("module.hdf5");
```
stevenj, line 37: As per the design documentation, the file name is to come from the manifest.
And is to have a `.hmod` extension.

I have reconsidered the idea of just copying everything from a directory and I don't like it.
Refer to the design documentation here:
https://input-output-hk.github.io/hermes/architecture/08_concepts/hermes_packaging_requirements/wasm_modules/#overview-of-a-wasm-component-module
The original design is that the manifest file controls how the package is built and is the source of the actual `metadata.json` file.  It is a superset of `metadata.json`.
The packaging process reads the manifest, collects the necessary file and produces the correctly named module.

I think the CLI should just be:
```
hermes package module <manifest_file> [<optional output path>]
```
`package` should have a sub command `module` because we will have multiple `packaging` type commands.
The manifest file is mandatory, and if the output path is not set, the output goes to the current working directory.
We shouldn't need to say `--manifest <file>`  it seems pointless if we always need the parameter, we should use positional arguments.  Same with output path.  `--something` type parameters should be reserved for things which are truly optional.

For now, rename this `.hdf5` extension to `.hmod`
The next PR will correct this, and use the manifest file as originally intended.
It will also then construct the `metadata.json` file not take one from the user, as all the data needed for `metadata.json` is either in the manifest file itself, or is generated by the build process.

hermes/bin/src/packaging/wasm_module.rs
```diff
@@ -0,0 +1,115 @@
...
+        let package = hdf5::File::create(&package_path)?;
+
```
stevenj, line 39: I wan't to see blosc compression set up properly and used on this file, i do not want it "left for later", because its the sort of thing that always gets delayed and then never done.  Its much easier to get it working with this simple builder than to try and wedge in and test later.  We have no reason to not use compressed images.  The necessary build options for HDF5 are added above, so it should work OK.

stevenj (changes requested):
See comments.
Changes to support the manifest file properly can be in the next PR.

I do require changes for the file extension, using blosc compression and the build using a static hdf5 library and properly enabling the compression filters though.

hermes/bin/src/cli/mod.rs
```diff
@@ -21,19 +31,45 @@ pub(crate) struct Cli {
...
+            .with_file(true)
+            .with_line_num(true)
+            .build();
+
+        logger::init(&log_config).unwrap_or_else(|err| errors.add_err(err));
+
         match self.command {
             None => run::Run::exec(),
-            Some(Commands::Package) => todo!(),
+            Some(Commands::Package(cmd)) => cmd.exec(),
```
stevenj, line 62: The command should be `hermes module ...`  `hermes package ...` should be for full application packages, not wasm modules.

stevenj: We should also stub the `hermes module sign` and `hermes module inspect` commands with their parameters in this PR, but those commands can just return and not do any actual work in this PR other than defining them.

hermes/bin/src/packaging/wasm_module.rs
Mr-Leshiy: I'm going to do it in the separate PR.
Next one

hermes/bin/src/packaging/wasm_module.rs
Mr-Leshiy: Updated

stevenj (approved):
LGTM

PR #240, opened 2024-05-24
feat: Add basic hdf5 compression

Mr-Leshiy:
# Description

Added basic compression implementation for hdf5 packages.

## Related Issue(s)

#237

### DIFF
--- a/.config/dictionaries/project.dic
+++ b/.config/dictionaries/project.dic
@@ -29,6 +29,7 @@ ciphertext
 codegen
 codepoints
 coti
+cpus
 crontabs
 crontagged
 cryptoxide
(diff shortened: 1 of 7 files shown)


hermes/bin/src/packaging/large.json
stevenj, line 1: Rather than adding a large test file to the git repo,  why don't we just store all the files in the project from the repo root?
This is also a large set of files of variable types, and you already had a function to store a directory into the HDF5 file.
As its mostly src code, it should have very good compression characteristics.

stevenj (changes requested):
My only concern is in the test file.
I think we can use the repo itself to give us an equivalent test data set.

hermes/bin/src/packaging/large.json
Mr-Leshiy: yea, will try that

Mr-Leshiy: updated

stevenj (dismissed):
LGTM

stevenj (approved):

PR #248, opened 2024-06-06
[WIP] test(sqlite): WASM integration test module

### LINKED ISSUE #246: 🛠️ [TASK] : Automated Test Cases running in CI.
### Summary

Create integration test running on WASM for the SQLite module.

### Details

Add an integration test module for SQLite in `wasm/integration-test` with the new directory `sqlite`. The test will be implemented with Rust bindings, and add the new test module to the CI. The test should comprise of decent cases to prove that the module can execute successfully on the WASM engine, and with decent benchmarks to compare querying performance between persistent (local file) and in-memory. Any encountered issues during writing the test to the module should be addressed.

apskhem:
# Description

Thanks for contributing to the project!
Please fill out this template to help us review your changes.

## Related Issue(s)

Closes #246

## Description of Changes

### DIFF
--- a/hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
+++ b/hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
@@ -1,3 +1,5 @@
+// cspell: words mapref
+
 //! Internal state implementation for the `SQLite` module.
 
 use std::collections::HashMap;
--- a/hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
+++ b/hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
@@ -1,0 +1,106 @@
+//! Internal state implementation for the `SQLite` module.
+
+use std::collections::HashMap;
+
+use dashmap::{mapref::one::RefMut, DashMap};
+use once_cell::sync::Lazy;
+
+use crate::app::HermesAppName;
+
+/// Represents an individual state for a particular object.
+#[derive(Debug)]
+pub(crate) struct ResourceObjectState {
+    /// A map holding key-value pairs of an object ID and a value.
+    id_map: HashMap<u32, usize>,
+    /// The current incremental state of ID.
+    current_id: Option<u32>,
+}
+
+/// Represents the state of resources.
+pub(crate) struct ResourceState {
+    /// The state of database object.
+    db_ptr_state: ResourceObjectState,
+    /// The state of database statement object.
+    stmt_ptr_state: ResourceObjectState,
+}
+
+impl ResourceObjectState {
+    /// Create a new `ResourceObjectState` with initial state.
+    fn new() -> Self {
+        Self {
+            id_map: HashMap::new(),
+            current_id: None,
+        }
+    }
+
+    /// Adds a value into the resource. If it does not exist, allocate one and returns the
+    /// new created key ID. In case of the key ID is running out of numbers, returns
+    /// `None`.
+    pub(super) fn allocate_object(&mut self, item_ptr: usize) -> Option<u32> {
+        if let Some((existing_id, _)) = self.id_map.iter().find(|(_, val)| val == &&item_ptr) {
+            Some(*existing_id)
+        } else {
+            let (new_id, is_overflow) = self
+                .current_id
+                .map_or_else(|| (0, false), |id| id.overflowing_add(1));
+
+            if is_overflow {
+                None
+            } else {
+                self.id_map.insert(new_id, item_ptr);
+                self.current_id = Some(new_id);
+                Some(new_id)
+            }
+        }
+    }
+
+    /// Retrieves a value according to its key ID.
+    pub(super) fn get_object_by_id(&self, id: u32) -> Option<usize> {
+        self.id_map.get(&id).map(ToOwned::to_owned)
+    }
+
+    /// Deletes a value according to its key ID, and returns the removed value if exists.
+    pub(super) fn delete_object_by_id(&mut self, id: u32) -> Option<usize> {
+        self.id_map.remove(&id)
+    }
+}
+
+impl ResourceState {
+    /// Create a new `ResourceState` with initial state.
+    pub(super) fn new() -> Self {
+        Self {
+            db_ptr_state: ResourceObjectState::new(),
+            stmt_ptr_state: ResourceObjectState::new(),
+        }
+    }
+
+    /// Gets the state for managing database objects.
+    pub(super) fn get_db_state(&mut self) -> &mut ResourceObjectState {
+        &mut self.db_ptr_state
+    }
+
+    /// Gets the state for managing statement objects.
+    pub(super) fn get_stmt_state(&mut self) -> &mut ResourceObjectState {
+        &mut self.stmt_ptr_state
+    }
+}
+
+/// Map of app name to resource holder
+type State = DashMap<HermesAppName, ResourceState>;
+
+/// Global state to hold `SQLite` resources.
+static SQLITE_INTERNAL_STATE: Lazy<State> = Lazy::new(DashMap::new);
+
+/// Represents the internal state object for `SQLite` module.
+pub(crate) struct InternalState;
+
+impl InternalState {
+    /// Set the state according to the app context.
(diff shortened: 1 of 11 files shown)


wasm/integration-test/sqlite/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+
+struct TestItem {
+    name: &'static str,
+    executor: fn() -> bool,
+}
+
+const TESTS: &[TestItem] = &[
+    TestItem {
+        name: "open-database-persistent-simple",
+        executor: || {
```
stevenj, line 27: These inline functions are quite long, it would be nicer if they were actual functions we just reference here by name.

stevenj (changes requested):
Can you also add preliminary results from your benchmarks, comparing the speed of the in-memory sqlite db and the on-disk one?

Add a readme to the integration-test directory with a summary of the results of the benchmark.  Make sure to include details of the machine it was run on.  This way we have something to compare it to if we run it somewhere else.

wasm/integration-test/sqlite/src/lib.rs
stevenj: Also, put the functions in another file, as this one is getting long already.

hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
```diff
@@ -0,0 +1,108 @@
...
+        Self {
+            id_map: HashMap::new(),
+            current_id: None,
+        }
+    }
+
+    /// Adds a value into the resource. If it does not exist, allocate one and returns the
+    /// new created key ID. In case of the key ID is running out of numbers, returns
+    /// `None`.
+    pub(super) fn allocate_object(&mut self, object_ptr: usize) -> Option<u32> {
```
Mr-Leshiy, line 41: Not a correct name for the function, because you are not actually allocating something init.
So it could mislead in the future.

hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::collections::HashMap;
+
+use dashmap::{mapref::one::RefMut, DashMap};
+use once_cell::sync::Lazy;
+
+use crate::app::HermesAppName;
+
+/// Represents an individual state for a particular object.
+#[derive(Debug)]
+pub(crate) struct ResourceObjectState {
```
Mr-Leshiy, line 14: This struct should be parametrised by the resource type.
With current implementation it's easy to mess taking an instance of the `db_state` for statement objects and broke an entire state.

hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// A map holding key-value pairs of an object ID and a value.
+    id_map: HashMap<u32, usize>,
+    /// The current incremental state of ID.
+    current_id: Option<u32>,
```
Mr-Leshiy, line 18: What is the purpose of this id ?

Mr-Leshiy (changes requested):
Seems it's time add a generalised object which will maintain `wasmtime::Resources`.
So this abstraction will be used in both `crypto` and `sqllite` runtime extensions.
It could be something like:
```Rust
trait ResourceAllocation {
   fn allocate<T>(params: T) -> Self {
    }
}

trait ResourceDeallocation {
   fn deallocate<T>(params: T) -> Self {
    }
}

struct ResourceManager<T> {
  apps_state: DashMap<HermesAppName, DashMap<u32, wasmtime::Resources<T>>>,
}

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs (hunk 2 above)
Comment (apskhem) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs:41]: I'll change this to `add_object`, should be a more straightforward name for this function. As I also think that `allocate_object` is more related about low-level memory stuffs.

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs (hunk 3 above)
Comment (apskhem) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs:14]: I can at least parameterized by specifying `<T: ToOwned + PartialEq>`.
So I can use as `ResourceObjectState<usize>`.

But since I tried to parameterized it by `ResourceObjectState<*mut sqlite3>`, it gave me this error...

`*mut sqlite3` cannot be sent between threads safely
within `(u32, *mut sqlite3)`, the trait `Send` is not implemented for `*mut sqlite3`

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs (hunk 4 above)
Comment (apskhem) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs:18]: It's an ID to work with `wasmtime::component::Resource` as `Resource` seems to need an internal ID to work with went creating a new resource object. An incremental ID works pretty fine on this regardless of its caveat that I don't know.

**Comment by apskhem** [comment]
> Seems it's time add a generalised object which will maintain `wasmtime::Resources`. So this abstraction will be used in both `crypto` and `sqllite` runtime extensions. It could be something like:
> 
> ```rust
> trait ResourceAllocation {
>    fn allocate<T>(params: T) -> Self {
>     }
> }
> 
> trait ResourceDeallocation {
>    fn deallocate<T>(params: T) -> Self {
>     }
> }
> 
> struct ResourceManager<T> {
>   apps_state: DashMap<HermesAppName, DashMap<u32, wasmtime::Resources<T>>>,
> }
> ```

Generalizing this is a great idea, but it might be a bit out of scope at the moment. Once the module is stable, we can revisit the idea of creating traits and structs for internal generalization with a dedicated ticket.

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs (hunk 2 above)
Comment (stevenj) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs]: I would rather we don't start doing this work for now.  its not necessary and just creates more work when we will probably need to refactor state management at a later date.  The simple implementation ehre is fine for now.

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs (hunk 3 above)
Comment (stevenj) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs]: I want a concrete example of the problem you are suggesting we have here @Mr-Leshiy ??

Comment (Mr-Leshiy) [reply hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs]: Already discussed it in private conversation.

File: hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs [hunk 5]
```diff
@@ -0,0 +1,108 @@
...
+use std::collections::HashMap;
+
+use dashmap::{mapref::one::RefMut, DashMap};
+use once_cell::sync::Lazy;
+
+use crate::app::HermesAppName;
+
+/// Represents an individual state for a particular object.
+#[derive(Debug)]
+pub(crate) struct ResourceObjectState<T: ToOwned + PartialEq> {
```
Mr-Leshiy, line 14: I think it's not necessary to make a generic this struct right now.
I think it would be enough just to provide a type alias for `usize`, something like this
```Rust
type Pointer = usize;
```

hermes/bin/src/runtime_extensions/hermes/sqlite/state.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub(super) fn get_stmt_state(&mut self) -> &mut ResourceObjectState<usize> {
+        &mut self.stmt_state
+    }
+}
+
+/// Map of app name to resource holder
+type State = DashMap<HermesAppName, ResourceState>;
+
+/// Global state to hold `SQLite` resources.
+static SQLITE_INTERNAL_STATE: Lazy<State> = Lazy::new(DashMap::new);
```
Mr-Leshiy, line 94:
```suggestion
static SQLITE_INTERNAL_STATE: Lazy<State> = Lazy::new(State::new);
```

Mr-Leshiy (dismissed):

stevenj (dismissed):

stevenj (approved):
LGTM

PR #253, opened 2024-06-13
feat(http gateway)

cong-or:
Initial HTTP gateway scaffold.

Future PRs which expand on this:
- extension of connection manager to include event tracking.
- example hello world kv app with defined API surface and wasm component to showcase - "how everything works in tandem"


hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -0,0 +1,168 @@
...
+        on_http_event,
+        TargetApp::All,
+        TargetModule::All,
+    ))?;
+
+    match &receiver.recv()? {
+        HTTPEventMsg::HttpEventResponseSome(resp) => {
+            Ok(Response::new(serde_json::to_string(&resp)?.into()))
+        },
+        HTTPEventMsg::HttpEventResponseNone() => Ok(not_found()),
```
cong-or, line 164: @stevenj in the case where the event responds with nothing (none). What should we return to the client? Should we classify nothing returned as a 404?

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -0,0 +1,168 @@
...
+    connection_manager
+        .connection_context
+        .try_lock()
+        .map_err(|_| anyhow::anyhow!("Unable to obtain mutex lock"))?
+        .insert(
+            unique_request_id,
+            (ClientIPAddr(ip), Processed(true), LiveConnection(false)),
+        );
+
+    info!("connection manager {:?}\n", connection_manager);
```
Mr-Leshiy, line 103:
```suggestion
    info!("connection manager {:?}", connection_manager);
```

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    for (header_name, header_val) in req.headers() {
+        header_map
+            .entry(header_name.to_string())
+            .or_default()
+            .push(header_val.to_str()?.to_string());
+    }
+
+    match uri.path() {
+        "/api" => {
```
Mr-Leshiy, line 127: I think `"api"` value should be as a const value

Mr-Leshiy (changes requested):
Am I correct that mechanics of this http extensions is follows right now:
http runtime extensions receives request -> for the received request transformed to hermes event -> hermes http event is sent to event queue -> hermes event executes inside wasm -> original request sender receives response after the successful event execution as a part of the event processing function.

What happens if event will be not executed inside wasm (because event queue does not provide such guaranties). Does the user will receive some response at this case ?

cong-or:
> Am I correct that mechanics of this http extensions is follows right now: http runtime extensions receives request -> for the received request transformed to hermes event -> hermes http event is sent to event queue -> hermes event executes inside wasm -> original request sender receives response after the successful event execution as a part of the event processing function.
> 
> What happens if event will be not executed inside wasm (because event queue does not provide such guaranties). Does the user will receive some response at this case ?

yes correct. But the next step is to add event tracking and management - the connection manager is a start for this. We will think about it in more detail then. At the moment we just respond [like this ](https://github.com/input-output-hk/hermes/blob/feature-http-gateway/hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs#L160) - the event either returns some or None and the http response mirrors this back to the client.

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
stevenj: We should not return anyting, UNTIL all wasm modules have finished.  At which time if none of them supplied a response we return 404.

cong-or:
re-opening with dependency tracking changes.

PR #267, opened 2024-06-27
feat: Implement self-referencing MultiEra block struct

### LINKED ISSUE #265: 🛠️ [TASK] : Decode a Cardano block uniformly and in a single place
### Summary

Wrangle the lifetime issues with decoded cardano blocks

### Description

The Pallas library has a function to convert a raw block to a decoded block.
We need to decode the block to use or reason through it, and to ensure blockchain integrity.
Decoding the block is not cheap.
Pallas places a lifetime on the decoded block which causes issues with an implementation that decodes the block once and attempts to reuse the decoded block.

Block data once read from chain is effectively immutable.  Individual blocks don't change.  So there is no reason why a block can't just be decoded once and the decoded version be used everywhere.

Decoding the block can also trigger errors, which need to be handled everywhere, even though the raw block data may already have been validated.

This task is to see if the decoding of the block can be encapsulated in such a way that a single decode can be reused without triggering an explosion of lifetime annotations throughout all of the code which uses it.

It will make the code cleaner, simpler and less error-prone.

The task is time boxed to a day and a half, if it cannot be solved in that time, it is to be deferred.

Mr-Leshiy:
# Description

Implemented a `SelfReferencedMultiEraBlock` which holds an instance of the block raw data with the decoded instance of the `pallas::ledger::traverse::MultiEraBlock`.

## Related Issue(s)

Closes #265


minikin (approved):
lgtm

stevenj:
This work was manually merged into the destination branch

PR #268, opened 2024-06-28
feat: Providing a generalised `Package` struct 

Mr-Leshiy:
# Description

Provided a generalised `Package` struct which incapsulates all implementation related to the HDF5 package


hermes/bin/src/packaging/package/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+    let reader = package
+        .dataset(name)
+        .ok()
+        .map(|ds| ds.as_byte_reader())
+        .transpose()?;
+    Ok(reader)
+}
+
+/// Buffer size for hash calculation.
+/// 1024 * 1024 = 1MB.
```
minikin, line 160: Can we exceed the size of the buffer? And if so, what will happen?

minikin (approved):
Approved with a question.

hermes/bin/src/packaging/package/mod.rs
Mr-Leshiy: No, this constant is used to setup buffer vector, which is used internally inside `calculate_package_file_hash` with the `Reader::read` function.

Issue #270, opened 2024-06-30
🛠️ [TASK] : Hermes application package


Mr-Leshiy:
### Summary

Implement a necessary functionality for building and signing hermes application packages.

### Description

- [x] Add cli capabilities to generate an application package
- [x] Add an application package abstraction and corresponding functionality
- [x] Implement validation logic of the application package
- [x] Add cli capabilities for signing

PR #281, opened 2024-07-02
feat: Hermes app package building [part 1]

Mr-Leshiy:
# Description

Implement a basic app application building functionality, with copying `srv/www` and `srv/share` directories and copying `metadata.json` file.

## Related Issue(s)

#270 

## Description of Changes

- Refactored `Errors` struct
- Added `PackagePath` struct, update `Package` code.
- Generalised `Metadata` object, so it used in `WasmModulePackage` and `ApplicationPackage`.
- Modified app manifest and metadata schemes, moved `icon` property from metadata to manifest definition.
- Added basic implementation of building `ApplicationPackage`.


hermes/bin/src/packaging/metadata.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl<T: MetadataSchema> Metadata<T> {
+    /// Create `Metadata` from reader.
+    pub(crate) fn from_reader(reader: impl Read) -> anyhow::Result<Self> {
+        let schema_validator = SchemaValidator::from_str(T::METADATA_SCHEMA)?;
+        let json = schema_validator.deserialize_and_validate(reader)?;
+        Ok(Self {
+            json,
+            _phantom: std::marker::PhantomData,
```
cong-or, line 51: why do we need phantom data here?

hermes/bin/src/errors.rs
```diff
@@ -35,13 +35,19 @@ impl Errors {
...
+        match err.downcast::<Errors>() {
+            Ok(errs) => self.0.extend(errs.0),
+            Err(err) => self.0.push(err),
+        }
     }
 
-    /// Merge two `Errors`
-    pub(crate) fn merge(&mut self, other: Self) {
-        self.0.extend(other.0);
+    /// Return a closure that adds an error to the `Errors`
```
cong-or, line 47: what's happening here? We are just collecting all the errors?

hermes/bin/src/packaging/metadata.rs
Mr-Leshiy:
```rust
pub(crate) struct Metadata<T> {
    /// metadata JSON object.
    json: serde_json::Map<String, serde_json::Value>,
    /// `T` type marker.
    _phantom: std::marker::PhantomData<T>,
}
```

The `Metadata` itself holds just a general purpose ` json: serde_json::Map<String, serde_json::Value>,` as a field.
I've added generic `T`  so in the case where application has their own metadata object and wasm module has their own metadata object, to not mess-up, have a type safety and distinguish these two different metadata objects.

P.S. Important to add that I've also added the following trait 
```rust
pub(crate) trait MetadataSchema {
    /// Metadata schema JSON string definition.
    const METADATA_SCHEMA: &'static str;
}
```
which should be implement by the type `T` to instantiate a `Metadata<T>` value.
This trait is used to statically specify a json schema for the metadata object.

hermes/bin/src/errors.rs
Mr-Leshiy: I've added this function which returns a closure to simplify such code from this
```rust
let mut errors = Errors::new();
some_function().unwrap_or_else(|err| errors.add_err(err))
```
to this
```rust
some_function().unwrap_or_else(errors.get_add_err_fn())
```

Mr-Leshiy: Yea, `add_err` just adds an error into the `Vec`.
The idea of this type is to collect all possible errors into one collection and return a collection of errors, instead of early return on the first error appearance (of course if it is possible to do).
so for example in this lines of codes
```rust

        let mut errors = Errors::new();

        validate_and_write_icon(manifest, &package).unwrap_or_else(errors.get_add_err_fn());
        validate_and_write_metadata(manifest, build_time, package_name, &package)
            .unwrap_or_else(errors.get_add_err_fn());
        write_www_dir(manifest, &package).unwrap_or_else(errors.get_add_err_fn());
        write_share_dir(manifest, &package).unwrap_or_else(errors.get_add_err_fn());
```
we are not returning if ` validate_and_write_icon` fails, we collect error from this function and move forward and only at the end we are returning all collected errors 
```rust
errors.return_result(Self(package))
```

PR #303, opened 2024-07-19
feat: read write files with runtime ext

cong-or:
allow access to virtual file system in runtime context via ` hermes runtime context`

### DIFF
--- a/hermes/bin/src/runtime_context.rs
+++ b/hermes/bin/src/runtime_context.rs
@@ -1,9 +1,9 @@
 //! Hermes runtime context implementation.
 
-use crate::{app::HermesAppName, wasm::module::ModuleId};
+use crate::{app::HermesAppName, vfs::Vfs, wasm::module::ModuleId};
 
 /// Hermes Runtime Context. This is passed to the WASM runtime.
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug)]
 pub(crate) struct HermesRuntimeContext {
     /// Hermes application name
     app_name: HermesAppName,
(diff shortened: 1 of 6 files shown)


hermes/bin/src/vfs/mod.rs
```diff
@@ -2,13 +2,82 @@
...
     #[allow(dead_code)]
     root: hermes_hdf5::Dir,
+    // TODO: add permissions RWX
+}
+
+impl Vfs {
+    /// Reads in data in bytes, the number of which is specified by the caller,
+    /// from the hdf5 file and stores then into a buffer supplied by the calling process.
+    #[allow(dead_code)]
+    pub(crate) fn read(&self, path: Path) -> anyhow::Result<Vec<u8>, anyhow::Error> {
```
FelipeRosa, line 24: I think we need to also have a fn that returns the File object so we can keep it open for reading/writing. That's something the WASI filesystem implementation is going to need.

I'm wondering if this PR will include that?

cong-or: we need to add permission too

cong-or: not in this one, just a basic scaffold in order to read to static files from runtime extension (for demo). We can iterate.

hermes/bin/src/runtime_context.rs
```diff
@@ -16,18 +16,24 @@ pub(crate) struct HermesRuntimeContext {
 
     /// module's execution counter
     exc_counter: u32,
+
+    /// App Virtual file system
+    #[allow(dead_code)]
+    vfs: Option<Vfs>,
```
Mr-Leshiy, line 22: I think it shouldn't be optional.
I know that it complicates initialisation of the runtime context for integration tests, but we still need an even even there.

Mr-Leshiy (changes requested):

hermes/bin/src/runtime_extensions/hermes/integration_test/event.rs
```diff
@@ -85,6 +87,10 @@ pub fn execute_event(
 ) -> anyhow::Result<Option<TestResult>> {
     let app_name = HermesAppName("integration-test".to_owned());
 
+    let hermes_home_dir = Cli::hermes_home()?;
```
Mr-Leshiy, line 90: As it is an integration tests, I think we should use `TempDir` for storing `vsf` instances.

Mr-Leshiy (approved):
Despite one comment LGTM

Mr-Leshiy (approved):

Issue #308, opened 2024-07-24
🛠️ [TASK] : VFS bootstrapping


Mr-Leshiy:
### Summary

Implement a hermes virtual file system bootstrapping part

### Description

During bootstrap process, need to open and existing one bfs hdf5 file (if exists) corresponded to the running app or create a new one.
Setup a dir structure following architecture docs https://input-output-hk.github.io/hermes/architecture/08_concepts/hermes_packaging_requirements/overview/#the-full-application-filesystem-hierarchy.
Mount all necessary files from the application package to the vfs instance.

PR #312, opened 2024-07-30
feat: VFS permissions handling

Mr-Leshiy:
# Description

Adding Hermes virtual file system permissions handling system.
The permissions state struct is based on the Radix tree structure, where each node is a path element with the corresponding permission level value.
By default each file and directory has a `ReadAndWrite` permission level.

## Related Issue(s)

#308 

## Description of Changes

- Updated `Vfs` and `VfsBotstrapper` structs, added permissions setup and usage.
- Added new `PermissionsState` and `PermissionLevel` types.
- Updated `AppPackage::mount_to_vfs` function with adding proper values of permission levels.

### DIFF
--- a/hermes/bin/src/app.rs
+++ b/hermes/bin/src/app.rs
@@ -1,6 +1,6 @@
 //! Hermes app implementation.
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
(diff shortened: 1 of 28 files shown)


cong-or (approved):
lgtm

PR #313, opened 2024-07-30
feat: initial regex for input validation and sanitization

cong-or:
**HTTP Gateway:**
Initial scaffold for regex input validation / sanitation and serving static files via VFS.


hermes/bin/src/runtime_extensions/hermes/http_gateway/mod.rs
```diff
@@ -1,13 +1,20 @@
...
 
+use crate::vfs::Vfs;
+
 mod event;
 mod gateway_task;
 /// Gateway routing logic
 mod routing;
 
+/// Virtual file system
+static VFS: OnceLock<Vfs> = OnceLock::new();
```
Mr-Leshiy, line 16: No, seems it's not correct.
As we dont have a global `Vfs` instance for all apps and each app has its own unique `Vfs` instance.

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -174,3 +180,64 @@ fn compose_http_event(
...
+        .get()
+        .ok_or(anyhow::anyhow!("Unable to obtain virtual filesystem"))?;
+
+    let file = vfs.read(path)?;
+
+    Ok(Response::new(file.into()))
+}
+
+/// Check if valid path to static files.
+fn is_valid_path(path: &str) -> anyhow::Result<bool> {
```
Mr-Leshiy, line 196: What do you think to even change the return type of this function to this.
So in the case of the invalid path error would be returned
```suggestion
fn is_valid_path(path: &str) -> anyhow::Result<()> {
```

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -174,3 +180,64 @@ fn compose_http_event(
...
+}
+
+#[cfg(test)]
+mod tests {
+    use regex::Regex;
+
+    use super::VALID_PATH;
+
+    #[test]
+    fn test_valid_paths_regex() {
```
Mr-Leshiy, line 209: I think inside this test we should cover `is_valid_path` function

Mr-Leshiy (approved):
What do you think, does it make sense to modify also `compose_http_event` with adding app_name as an argument ?
So inside it you could already trigger event to the specific hermes app not to all of them
LGTM.

cong-or:
> What do you think, does it make sense to modify also `compose_http_event` with adding app_name as an argument ? So inside it you could already trigger event to the specific hermes app not to all of them LGTM.

This is a topic we need to discuss further, I think there is an assumption that the event goes to all of them.

PR #322, opened 2024-08-03
refactor: Hermes Reactor

Mr-Leshiy:
# Description

Refactored hermes reactor application's state management.
Made a reactor state as global static which is internally managed by the reactor's module.
So all other modules as event queue could get a state of currently available apps by calling public functions of the reactor module.

## Description of Changes

- Renamed `HermesReactor` to `Reactor`, `HermesApp` to `Application`, `HermesAppName` to `ApplicationName`.
- Added new reactor api with `init`, `load_app`, `get_app`, `get_all_app_names` free functions.
- Updated event queue implementation in corresponds with the new reactor.
- Updated http gateway `serve_static_data` function with the usage of `reactor::get_app` function. So it is not needed anymore to maintain `vfs` state internally.
- Refactored cli commands prints, moved application building process to the new `packaging::app::app_builder::build_app` function.


cong-or (dismissed):
lgtm

cong-or (approved):
lgtm

PR #323, opened 2024-08-05
feat: Bump cat-ci version, cleanup Rust tests

Mr-Leshiy:
# Description

- Bumped cat-ci version
- Cleanup rust tests and clippy linting


hermes/bin/src/hdf5/compression.rs
```diff
@@ -92,27 +92,19 @@ mod tests {
     /// it and checks the difference.
     #[test]
     fn blosc_compression_test() {
-        let dir = TempDir::new().expect("cannot create temp dir");
+        let dir = TempDir::new().unwrap();
```
stevenj, line 95: `expect` should be ok for tests, why change to unwrap()?  They both panic.

stevenj (changes requested):

hermes/bin/src/hdf5/compression.rs
Mr-Leshiy: Just to make code cleaner, as this expect message does not provide additional context. The same for others.

kukkok3: I would leave the expect and use a better error message it that doesnt provide enough context. Otherwhise when the unwap panics it is really difficult to understand what's going on

Mr-Leshiy: But that's become a problem and a headache rather than improving error messaging itself (what is actually should be done).
I agree that in some specific places in some special cases we can use `expect` with some message to provide more context. But it's not the case of this refactoring, originally all these `expects` was placed just because `unwrap` was not allowed and most of the messages just does not make much sense.
So I would rather follow that in most of the cases use `unwrap` and in some special cases use `expect`

kukkok3: Why it has become a problem? 
The unwrap was not allowed for this exact reason (when did we allowed it again? I think I missed that), we wanted to force the use of expect because it is really difficult to track down the source of the panic otherwhise and I dont think we want to implement proper error handling for the tests. If the messages dont make sense we should replace them with some messages that make sense, not remove the expect imho :)
I also would suggest to not allow the unwrap again.
Quoting the [rust book](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html)

> Similarly, the expect method lets us also choose the panic! error message. Using expect instead of unwrap and providing good error messages can convey your intent and make tracking down the source of a panic easier.

Mr-Leshiy: >Why it has become a problem?
Not a problem actually, mainly a headache.
For example with this line
```
let dir = TempDir::new().unwrap();
```
We are trying to create a temporary directory with the help of the `tmp_dir` crate. If we are getting an error, basically it means that we cannot create it for some reason, and that's obvious and do not require any additional context. Adding any `expect` msg here will not actually help you to solve anything. To solve the potential problem you need to dig into what exact error `TempDir::new` returns to you (and that's will showed with unwrap).
As a result right now we have a tons of same  `.expect("cannot create temp dir")` all around the code which does not add any additional value and in other common places where.
And it becomes a headache because when you are writing some unit test like this, you are trying to think what `expect` message I need to add for actually obvious error cases. And when you are adding it everywhere, code becomes blowed up with a lot of messages which at the end does not mean nothing to you really.
```Rust
let share_dir = package
            .0
            .create_dir(ModulePackage::SHARE_DIR.into())
            .unwrap();
        let new_file_name = "new_file";
        let new_file_content = b"new file content";
        let mut new_file = share_dir.create_file(new_file_name.into()).unwrap();
        new_file.write_all(new_file_content).unwrap();
        assert_ne!(module_package_content.share.file.0.as_str(), new_file_name);
        assert_ne!(
            module_package_content.share.file.1.as_slice(),
            new_file_content
        );
```

> because it is really difficult to track down the source of the panic and I dont think we want to implement proper error handling for the tests
As we are testing some our objects, functions api, we must keep error handling clear because at the end it will be propagated in production and could be visible for users or for us as a `DEBUG` log message. So we definitely want to keep error handling. It's not just for tests, it's for users and for us.
If we want to make error from the function more understandable, we need to fix our error message, not by providing more context only on the test side.

Mr-Leshiy: Just believe me, after I've written few unit tests for packaging module, this "must to use `expect`" rule, was really starting to annoy, because you just dont know what to add more despite that what is already expressed with the code.
 And seems I'm not the only one.

kukkok3: @Mr-Leshiy I mean I understand your point, but also believe me I have spent a lot of time tracking down the source of a panic when a test fails...maybe something like `expect("cannot create temp dir in + {name of the test}");` would be better? but yeah let's discuss on friday!

Mr-Leshiy: On our tech discussion decided to enable backtracing, which was done in this PR https://github.com/input-output-hk/catalyst-ci/pull/289

kukkok3 (approved):
LGTM

stevenj (approved):
LGTM

PR #425, opened 2025-06-25
fix(wasm): Update Cardano Chain Follower WIT interface

bkioshn:
# Description

Modify the current WIT interface of Cardano, specificaly Cardano Chain Follower.

## Related Issue(s)

Closed #419

## Description of Changes

Import - Implement by Host
- block resource
- network resource
- transaction resource
- subscription id resoruce

Export - An event which Implement by WASM component
- on-cardano-block()
- on-cardano-roll-forward() 

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [ ] My changes generate no new warnings
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] New and existing unit tests pass locally with my changes
* [ ] Any dependent changes have been merged and published in downstream module

Issue #420, opened 2025-06-09
🛠️ [TASK] : Update Runtime Extension for Chain Follower


stevenj:
### Summary

Runtime Extension needs to be updated to match the new WIT interface

### Description

Update the Rust code in hermes to use the catalyst-libs cardano chain follower, and remove the chain follower in hermes.

PR #436, opened 2025-07-08
fix(hermes): New Cardano runtime implementation

bkioshn:
# Description

Update the Cardano runtime implementation to match the new WIT interface
https://github.com/input-output-hk/hermes/pull/425

## Related Issue(s)
#420 

## Breaking Changes

A new interface has been defined and implemented to replace the previous behavior.

## Related Pull Requests

The WIT interface is defined here
https://github.com/input-output-hk/hermes/pull/425


hermes/bin/src/runtime_extensions/hermes/cardano/event.rs
```diff
@@ -1,20 +1,29 @@
...
-    /// This raw CBOR block data.
-    pub(super) block: CardanoBlock,
-    /// Source information about where the block came from, and if we are at tip or not.
-    pub(super) source: BlockSrc,
+    /// The Cardano blockchain network that the block event is originated from.
+    network: CardanoNetwork,
+    /// A unique identifier of the block subscription.
+    subscription_id: SubscriptionId,
+    /// The Cardano slot number that the block event is in.
+    slot: Slot,
```
stevenj, line 20: see comment on wit PR.  I think this should just be the block resource, and all of this other data can be obtained directly from that resource, so it doesn't need to be in the event directly.

hermes/bin/src/runtime_extensions/hermes/cardano/event.rs
```diff
@@ -1,20 +1,29 @@
...
-    /// The blockchain id the block originated from.
-    pub(super) blockchain: CardanoBlockchainId,
-    /// This raw CBOR block data.
-    pub(super) block: CardanoBlock,
-    /// Source information about where the block came from, and if we are at tip or not.
-    pub(super) source: BlockSrc,
+    /// The Cardano blockchain network that the block event is originated from.
+    network: CardanoNetwork,
+    /// A unique identifier of the block subscription.
+    subscription_id: SubscriptionId,
```
stevenj, line 18: I am still thinking, maybe SubscriptionId should be a resource, with a `network` and `unsubscribe` methods attached to it?

hermes/bin/src/runtime_extensions/hermes/cardano/event.rs
```diff
@@ -1,20 +1,29 @@
...
+    /// The Cardano slot number that the block event is in.
+    slot: Slot,
+    /// The CBOR format of the Cardano block data.
+    block: Vec<u8>,
+    /// Flag indicate whether the block is immutable or not.
+    is_immutable: bool,
+    /// Flag indicate whether the block is a rollback.
+    is_rollback: bool,
```
stevenj, line 26: With the above two changes, the event would just be:
```
pub(super) struct OnCardanoBlockEvent {
    subscription_id: SubscriptionId,
    block: Block
}
```
which looks a lot cleaner / easier to manage / maintain.

stevenj (changes requested):

hermes/bin/src/runtime_extensions/hermes/cardano/event.rs
bkioshn: Will pass`subscription_id: u32`  and `block: u32` where the u32 represent the resource representation. 
Cant directly clone the resource since it is not implemented. So will construct a borrow version of these resource and pass to wasm.

PR #441, opened 2025-07-14
feat(hermes): Add svg validation

### LINKED ISSUE #282: 🛠️ [TASK] :  Add `svg` files validation during package building
### Summary

Add s simple `svg` files validation, that this file actually an svg image.

### Description

- Added svg files validation by founding an appropriate crate for doing it.
- Update app loading functionality, `fn validate_and_write_icon` with svg validation.

bkioshn:
# Description

Add package validation

## Related Issue(s)

Closes #282 


hermes/bin/src/packaging/app/mod.rs
```diff
@@ -359,11 +359,28 @@ impl ApplicationPackage {
...
+    if !path.to_string().ends_with("svg") {
+        return Ok(false);
+    }
+
+    let mut reader = resource.get_reader()?;
+    let mut buf = Vec::new();
+    std::io::copy(&mut reader, &mut buf)?;
+
+    let opt = usvg::Options::default();
+    Ok(usvg::Tree::from_data(&buf, &opt).is_ok())
```
rafal-ch, line 381: What do you think about introducing some more security measures? While `from_data()` protects us from accepting malformed files, we are still going to accept files that are unnecessarily huge or require suspiciously long processing times (see: "Billion laughs attack" for XML parsers).

Please consider adding:
1. some size limit (I think 100-200kb should be good for most cases)
2. timeout on processing (if `from_data()` takes more than, let's say 500ms) we should error and not accept the svg

bkioshn: - usvg is already have this [ElementsLimitReached](https://docs.rs/usvg/latest/usvg/enum.Error.html#variant.ElementsLimitReached) which I think should already prevent a billion laugh style attacks.
- I dont think we need a size limit due to svg files can vary depending on their use case and design complexity

rafal-ch: Appoved, but I'd still be in favor of a simple size check, so that we can fail if someone tries to package a 1Gb large .svg. But I'm not strong on this hence leaving the decision to you :)

rafal-ch (approved):

PR #446, opened 2025-07-18
Feat/http gateway proxy

cong-or:
### 🔄 Internal Redirect System
The redirect functionality implements a **secure internal redirect system** that allows WASM modules to safely redirect HTTP requests to other internal services while maintaining strict security controls.
#### How It Works
1. **WASM Module Initiation**: A WASM module processes an incoming HTTP request and determines it needs to redirect to another internal service
2. **WIT Interface Response**: The module returns an variant through the WebAssembly Component Interface, providing type-safe communication between WASM and host `internal-redirect(string)`
3. **Security Validation**: The redirect URL undergoes comprehensive validation against configurable security policies
4. **Asynchronous Processing**: The redirect request is executed in a background thread to avoid blocking the original request handler
5. **Response Forwarding**: The response from the redirect target is forwarded back to the original client

#### Security Model
The redirect system implements a **whitelist-based security model** with three validation layers:
**URL Scheme Validation**
- Only allows predefined schemes (default: `https`)
- Prevents protocol downgrade attacks
- Configurable via environment variable `REDIRECT_ALLOWED_SCHEMES`

**Host Validation**
- Restricts redirects to approved internal hosts
- Prevents external redirect attacks and data exfiltration
- Default: `app.dev.projectcatalyst.io`
- Configurable via environment variable `REDIRECT_ALLOWED_HOSTS`

**Path Prefix Validation**
- Ensures redirects only target approved API endpoints
- Prevents unauthorized access to internal services
- Default: `/api/gateway`
- Configurable via environment variable `REDIRECT_ALLOWED_PATH_PREFIXES`

#### Architecture Benefits
**Header Preservation**: Maintains original request headers (authentication, tracing, etc.) when forwarding to internal services 
**Non-blocking Design**: Uses background threads to prevent request timeouts during internal service calls 
**Error Isolation**: Failed redirects don't crash the gateway - errors are logged and appropriate HTTP error responses are returned

### DIFF
--- a/hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
+++ b/hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
@@ -3,17 +3,17 @@
 //! Processes HTTP requests through WASM modules using MPSC channels.
 //! Validates redirect URLs against configurable security policies.
 
-use std::collections::HashSet;
-use std::env;
-use std::sync::mpsc::Sender;
-use tracing::error;
+use std::{collections::HashSet, env, result::Result::Ok, sync::mpsc::Sender};
 
 use hyper::{self, body::Bytes};
 use serde::{Deserialize, Serialize};
+use tracing::error;
 use url::Url;
 
-use crate::event::HermesEventPayload;
-use crate::runtime_extensions::bindings::exports::hermes::http_gateway::event::HttpGatewayResponse;
+use crate::{
+    event::HermesEventPayload,
+    runtime_extensions::bindings::exports::hermes::http_gateway::event::HttpGatewayResponse,
+};
 
 // ============================================================================
 // Type Aliases
@@ -43,7 +43,7 @@ type Body = Vec<u8>;
 pub(crate) enum HTTPEventMsg {
     /// Receiver acknowledgment
     HTTPEventReceiver,
-    /// Event response: (status_code, headers, body)
+    /// Event response: (`status_code`, `headers`, `body`)
     HttpEventResponse((Code, HeadersKV, Body)),
 }
 
@@ -53,9 +53,13 @@ pub(crate) enum HTTPEventMsg {
 
 /// HTTP request event to be processed by WASM modules
 pub(crate) struct HTTPEvent {
+    /// HTTP headers as key-value pairs
     pub(crate) headers: HeadersKV,
+    /// HTTP method (GET, POST, etc.)
     pub(crate) method: Method,
+    /// Request URL path
     pub(crate) path: Path,
+    /// Request body content
     pub(crate) body: Bytes,
     /// Channel to send response back to client
     pub(crate) sender: Sender<HTTPEventMsg>,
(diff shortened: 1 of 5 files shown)


cong-or:
p.s the environment variable approach is temporary until we design a better configuration strategy.

hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
```diff
@@ -60,14 +203,137 @@ impl HermesEventPayload for HTTPEvent {
...
+    }
+
+    /// Spawn background thread for redirect request
+    fn spawn_redirect_request(&self, location: String) {
+        let headers = self.headers.clone();
+        let method = self.method.clone();
+        let body = self.body.clone();
+        let sender = self.sender.clone();
+
+        std::thread::spawn(move || {
```
rafal-ch, line 246: We might consider issuing a spike to investigate if it is feasible to allow async in the event handlers, so we can use tokio task instead of OS thread.

hermes/bin/src/runtime_extensions/hermes/http_gateway/routing.rs
```diff
@@ -145,7 +145,7 @@ async fn route_to_hermes(
     }
 
     let (_parts, body) = req.into_parts();
-    if uri.path() == WEBASM_ROUTE {
+    if uri.path() == WEBASM_ROUTE || uri.path().starts_with(&format!("{WEBASM_ROUTE}/")) {
```
rafal-ch, line 148: I think we also need to take care of the case sensitivity here. The [docs](https://docs.rs/http/1.3.1/http/uri/struct.Uri.html#method.path) for `path()` say that _"The path component is case sensitive."_

hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
cong-or: yes I agree, it would improve performance.. I couldn't find a clean way of doing it so kept it simple for now.

hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
```diff
@@ -60,14 +203,137 @@ impl HermesEventPayload for HTTPEvent {
...
+    /// Send HTTP response back to client via MPSC channel
+    fn send_http_response(&self, code: Code, headers: HeadersKV, body: Body) -> anyhow::Result<()> {
+        Ok(self
+            .sender
+            .send(HTTPEventMsg::HttpEventResponse((code, headers, body)))?)
+    }
+
+    /// Handle internal redirect with security validation
+    fn handle_internal_redirect(&self, location: String) -> anyhow::Result<()> {
+        let config = RedirectConfig::from_env();
```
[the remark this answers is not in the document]
cong-or: agreed

hermes/bin/src/runtime_extensions/hermes/http_gateway/event.rs
```diff
@@ -1,51 +1,194 @@
...
+    validate_scheme(&url, config)?;
+    validate_host(&url, config)?;
+    validate_path(&url, config)?;
+
+    Ok(())
+}
+
+/// Validates URL scheme against allowed schemes
+fn validate_scheme(url: &Url, config: &RedirectConfig) -> anyhow::Result<()> {
+    if !config.schemes.contains(url.scheme()) {
```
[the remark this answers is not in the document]
cong-or: yes, I thought I flattened everything to lowercase but looks like I forgot it.

PR #448, opened 2025-07-18
feat(rust): Hermes done

### LINKED ISSUE #447: 🛠️ [TASK] : Hermes apps/modules should have a way of exiting with an exit code
### Summary

Hermes apps/modules should have a way of exiting with an exit code

### Description

This functionality is needed so that we can build a proper integration tests. The general idea is that we build and "integration test" app/module, execute it on hermes and check the exit code. This is needed because the current integration testing framework only allows to call and check the single function invocation (like a unit-test) and is not able to validate more complex behavior involving multiple events flowing between modules.

## Definition of done
There is a new "done(exitcode)" function. When it is invoked the hermes itself shuts down and reports the specified code to the operating system.

## Acceptance Criteria

- [ ] Add the `done(exitcode: int)`  function to the `init` wit module interface.
- [ ] When called, the node immediately exists and returns the exit code as its return code.

This is a Minimum implementation which will be expanded in future work, but gets us what we need immediately.

no30bit:
# Description

Implements support for shutting down Hermes.

## Related Issue(s)

Closes #447

## Description of Changes

- New `Exit` type descrIbing status of the Hermes shutdown.
- New `ExitLock` type. Can be used to wait until the event queue is shut down.
- Event queue now receives events wrapped in `ControlFlow::Continue`
  - Sending `ControlFlow::Break(exit_code)` to queue is used as a manual shutdown.
- Added `event::queue::shutdown(exit_code)` function.
- Updated CLI `run` command.
  - Now printing the exit status.
  - Replace `yield_now() + loop {}` with `Condvar` waiting (wrapped in `ExitLock`).
  - Add `--timeout-ms` argument.
- Add `done` function to `hermes-init` WIT world.
- Add  `Host` implementation for `done` that shuts down the node.

## Breaking Changes

Security is majorly hurt. After this PR any application would be able to shut down the whole node.
A solution would be to gate both `done` fn and `integration-test` world behind WIT feature(s) in WIT files.
These features can then be toggled in Hermes `LinkOptions` trivially.

### DIFF
--- a/hermes/bin/src/cli/mod.rs
+++ b/hermes/bin/src/cli/mod.rs
@@ -78,7 +78,10 @@ impl Cli {
         logger::init(&log_config).unwrap_or_else(errors.get_add_err_fn());
 
         match self.command {
-            Commands::Run(cmd) => cmd.exec(),
+            Commands::Run(cmd) => {
+                cmd.exec()
+                    .map(|exit| println!("{}:\n{}", Emoji::new("⛔", "Exit"), style(exit).red()))
+            },
             Commands::Module(cmd) => cmd.exec(),
             Commands::App(cmd) => cmd.exec(),
         }
--- a/hermes/bin/src/event/queue.rs
+++ b/hermes/bin/src/event/queue.rs
@@ -1,10 +1,14 @@
 //! Hermes event queue implementation.
 
+mod exit;
+
 use std::{
-    sync::mpsc::{Receiver, Sender},
+    ops::ControlFlow,
+    sync::mpsc::{self, Receiver, Sender},
     thread::{self},
 };
 
+pub use exit::{Exit, ExitCode, ExitLock};
 use once_cell::sync::OnceCell;
 
 use super::{HermesEvent, TargetApp, TargetModule};
@@ -17,6 +21,11 @@ static EVENT_QUEUE_INSTANCE: OnceCell<HermesEventQueue> = OnceCell::new();
 #[derive(thiserror::Error, Debug, Clone)]
 #[error("Failed to add event into the event queue. Event queue is closed.")]
 pub(crate) struct CannotAddEventError;
+
+/// Failed to add event into the event queue. Event queue is closed.
+#[derive(thiserror::Error, Debug, Clone)]
+#[error("Failed to shutdown event queue. Event queue is closed.")]
+pub(crate) struct CannotShutdownError;
 
 /// Failed when event queue already been initialized.
 #[derive(thiserror::Error, Debug, Clone)]
@@ -32,25 +41,29 @@ pub(crate) struct NotInitializedError;
 /// It is a singleton struct.
 struct HermesEventQueue {
     /// Hermes event queue sender
-    sender: Sender<HermesEvent>,
+    sender: Sender<ControlFlow<ExitCode, HermesEvent>>,
 }
 
 /// Creates a new instance of the `HermesEventQueue`.
 /// Runs an event loop thread.
 ///
+/// [`ExitLock`] would contain shutdown information if awaited.
+///
 /// # Errors:
 /// - `AlreadyInitializedError`
-pub(crate) fn init() -> anyhow::Result<()> {
+pub(crate) fn init() -> anyhow::Result<ExitLock> {
     let (sender, receiver) = std::sync::mpsc::channel();
 
     EVENT_QUEUE_INSTANCE
         .set(HermesEventQueue { sender })
         .map_err(|_| AlreadyInitializedError)?;
 
+    let (exit_tx, exit_rx) = ExitLock::new_pair();
     thread::spawn(move || {
-        event_execution_loop(receiver);
+        let exit = event_execution_loop(&receiver);
+        exit_tx.set(exit);
     });
-    Ok(())
+    Ok(exit_rx)
 }
 
 /// Add event into the event queue
@@ -61,7 +74,26 @@ pub(crate) fn init() -> anyhow::Result<()> {
 pub(crate) fn send(event: HermesEvent) -> anyhow::Result<()> {
     let queue = EVENT_QUEUE_INSTANCE.get().ok_or(NotInitializedError)?;
 
-    queue.sender.send(event).map_err(|_| CannotAddEventError)?;
+    queue
+        .sender
+        .send(ControlFlow::Continue(event))
+        .map_err(|_| CannotAddEventError)?;
+
+    Ok(())
+}
+
+/// Shutdown the event queue.
+///
+/// # Errors:
+/// - `CannotShutdownQueueError`
+/// - `NotInitializedError`
+pub(crate) fn shutdown(code: ExitCode) -> anyhow::Result<()> {
+    let queue = EVENT_QUEUE_INSTANCE.get().ok_or(NotInitializedError)?;
+
+    queue
+        .sender
+        .send(ControlFlow::Break(code))
+        .map_err(|_| CannotShutdownError)?;
 
     Ok(())
 }
@@ -109,9 +141,13 @@ fn targeted_app_event_execution(event: &HermesEvent) {
     }
 }
 
-/// Executes Hermes events from the provided receiver .
-fn event_execution_loop(receiver: Receiver<HermesEvent>) {
-    for event in receiver {
-        targeted_app_event_execution(&event);
+/// Executes Hermes events from the provided receiver.
+fn event_execution_loop(receiver: &Receiver<ControlFlow<ExitCode, HermesEvent>>) -> Exit {
+    loop {
+        match receiver.recv() {
+            Ok(ControlFlow::Continue(event)) => targeted_app_event_execution(&event),
+            Ok(ControlFlow::Break(exit_code)) => break Exit::Done { exit_code },
+            Err(mpsc::RecvError) => break Exit::QueueClosed,
+        }
     }
 }
--- a/hermes/bin/src/event/queue/exit.rs
+++ b/hermes/bin/src/event/queue/exit.rs
@@ -1,0 +1,95 @@
+//! Implementation of exit status retrieval after event queue shutdown.
+
+use std::{
+    sync::{Arc, Condvar, Mutex},
+    time::Duration,
+};
+
+/// Exit code to return after queue shutdown.
+pub type ExitCode = i64;
+
+/// Exit status.
+#[derive(Copy, Clone, PartialEq, Eq, Debug, thiserror::Error)]
+pub enum Exit {
+    /// An application requested runtime abort.
+    #[error(
+        "Event queue closed: an application requested runtime abort with exit code {exit_code}"
+    )]
+    Done {
+        /// Exit code provided by an application.
+        exit_code: ExitCode,
+    },
+    /// All event senders are dropped.
+    #[error("Event queue closed: all event senders are dropped")]
+    QueueClosed,
+    /// Another task panicked inside.
+    #[error("Event queue poisoned: another task panicked inside")]
+    QueuePoisoned,
+    /// Timeout elapsed.
+    #[error("Event queue closed: timeout")]
+    Timeout,
+}
+
+/// Lock to [`Exit`] that event queue sets on shutdown.
+// To satisfy the guarantees [`ExitLock`] should never implement or derive `Clone`.
+pub struct ExitLock(Arc<(Condvar, Mutex<Option<Exit>>)>);
+
+impl ExitLock {
+    /// Initializes the lock and return two clones of it.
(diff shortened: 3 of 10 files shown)


hermes/bin/src/cli/mod.rs
```diff
@@ -78,7 +78,10 @@ impl Cli {
         logger::init(&log_config).unwrap_or_else(errors.get_add_err_fn());
 
         match self.command {
-            Commands::Run(cmd) => cmd.exec(),
+            Commands::Run(cmd) => {
+                cmd.exec()
+                    .map(|exit| println!("{}:\n{}", Emoji::new("⛔", "Exit"), style(exit).red()))
```
stevenj, line 83: How is this propagated by the the processes return code?
So, i could use `make` or `bash` to check if the `hermes` command executed correctly or failed with an error, and what error code it was?

wasm/wasi/wit/deps/hermes-init/api.wit
```diff
@@ -0,0 +1,10 @@
+/// # Interface to Hermes runtime control flow.
+///
+/// ## Permissions
+///
+/// This API is ONLY Available to Integration Tests and is not normally exposed.
+
+interface api {
+    /// Perform Hermes event queue shut down.
+    done: func(exit-code: s64);
```
stevenj, line 9: To align with posix exit code requirements, this should be a u8, and the allowable values are 0-255.  
Windows allows u32, but we need to adopt the lowest common denominator here, so `u8`.

stevenj (changes requested):

hermes/bin/src/cli/mod.rs
no30bit: Right, I'll propagate it

hermes/bin/src/event/queue.rs
```diff
@@ -109,9 +141,13 @@ fn targeted_app_event_execution(event: &HermesEvent) {
...
 
-/// Executes Hermes events from the provided receiver .
-fn event_execution_loop(receiver: Receiver<HermesEvent>) {
-    for event in receiver {
-        targeted_app_event_execution(&event);
+/// Executes Hermes events from the provided receiver.
+fn event_execution_loop(receiver: &Receiver<ControlFlow<ExitCode, HermesEvent>>) -> Exit {
+    loop {
+        match receiver.recv() {
+            Ok(ControlFlow::Continue(event)) => targeted_app_event_execution(&event),
```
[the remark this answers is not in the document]
no30bit: If you don't propose to add `Done` event to WIT, then it would be inconsistent to make it implement `HermesEvent` in Rust. I went with this assumption while implementing this PR.

My reasoning is that `ControlFlow` is a part of the queue logic. It's  what reactor consumes.
While user consumes `HermesEvents` through callbacks. It's thus wrapped in reactor's `ControlFlow`.
Now, user has access to the reactor by either implicitly agreeing to `Continue` or explicitly issuing a `Break`. Not through the callback like with `HermesEvents`.

wasm/wasi/wit/deps/hermes-init/api.wit
no30bit: FIxed

hermes/bin/src/event/queue.rs
```diff
@@ -109,9 +141,13 @@ fn targeted_app_event_execution(event: &HermesEvent) {
...
+            Ok(ControlFlow::Break(exit_code)) => break Exit::Done { exit_code },
```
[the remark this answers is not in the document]
no30bit: >  single producer, multi consumer queue
> any event which could target multiple wasm modules would need to be expanded into multiple events

So, we would put 10 individual events into the queue rather than one targeting 10 modules?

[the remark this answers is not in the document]
stevenj: No, if the event targeted 10 modules, it would be 10 events into one multi consumer queue. Not sure that's the best way to do things, but it would work.  Each thread in the worker pool would be able to consume from the queue, and run a single module.  So whichever worker is ready first gets the oldest event to run.  This is just one way to do it, not necessarily the best or most performant.

wasm/wasi/wit/deps/hermes-init/api.wit
```diff
@@ -0,0 +1,13 @@
+/// # Interface to Hermes runtime control flow.
+///
+/// ## Permissions
+///
+/// This API is ONLY Available to Integration Tests and is not normally exposed.
+
+interface api {
+    /// Perform Hermes event queue shutdown.
+    done: func(status: result);
```
stevenj, line 9: I wasn't suggesting implementing all of the exit() just pointing to the status code needing to be a `u8`.
I think `done: func(status: u8)` is more than enough for our purposes.

hermes/bin/src/event/queue/exit.rs
```diff
@@ -0,0 +1,95 @@
...
+        // Default value doesn't matter as it's bound to be awaited to be changed.
+        let payload = Mutex::new(None);
+        let inner = Arc::new((condvar, payload));
+        (Self(inner.clone()), Self(inner))
+    }
+
+    /// Set the [`Exit`] value. This will notify the waiting thread.
+    ///
+    /// This shouldn't exposed the waiting thread.
+    pub(super) fn set(self, exit: Exit) {
```
cong-or, line 71: **ExitLock shutdown inefficiency** - When multiple threads call `shutdown()` at the same time, they all grab the lock and notify waiters even though only the first one actually sets the exit code. Should check if the value is already set before notifying to avoid unnecessary work

```rust
if payload.is_none() {
    *payload = Some(exit);
    condvar.notify_one();
}
```

hermes/bin/src/runtime_extensions/hermes/init/host.rs
```diff
@@ -0,0 +1,23 @@
...
+use std::process::ExitCode;
+
+use crate::{
+    event, runtime_context::HermesRuntimeContext,
+    runtime_extensions::bindings::hermes::init::api::Host,
+};
+
+impl Host for HermesRuntimeContext {
+    /// Perform Hermes event queue shut down.
+    fn done(&mut self, status: Result<(), ()>) -> wasmtime::Result<()> {
```
cong-or, line 11: Is it possible for Multiple WASM instances to call `done()` simultaneously, sending conflicting exit codes. Consider making shutdown idempotent with an `AtomicBool` guard to ensure only the first shutdown attempt succeeds.

hermes/bin/src/runtime_extensions/hermes/init/host.rs
```diff
@@ -0,0 +1,23 @@
...
+        if status.is_ok() {
```
cong-or, line 12: **Missing error propagation**: `done()` and `done_with_code()` ignore `shutdown()` errors and always return . Should propagate shutdown failures to the WASM runtime. `Ok(())`
**Impact**: Silent failures, resource leaks, and zombie processes when shutdown actually fails

hermes/bin/src/event/queue/exit.rs
stevenj: This would race if multiple threads can call at once.
Suggest using a atomic i16, set to negative if its not set, then you can use compare and set atomic operations to update it once and then notify.  Other updates will not notify, they will end.

no30bit: @stevenj 
Multiple threads don't call this.
Using `Condvar` is preferred in my opinion as it has great utility functions (i.e. `wait` and `wait_timeout`).
Implementing it with atomics isn't as trivial and/or optimal.

What I'd really want to use instead would be the [`OnceLock::wait`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.wait). However, on our rust version it's not available.

hermes/bin/src/runtime_extensions/hermes/init/host.rs
no30bit: This is already guaranteed. Event loop breaks at first `shutdown`.

no30bit: If it were to be guaranteed at this level too, would you agree with `static EXIT: OnceLock` implementation?

hermes/bin/src/runtime_extensions/hermes/init/host.rs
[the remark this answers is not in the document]
no30bit: From that commit
https://github.com/input-output-hk/hermes/pull/448/files/324992bc877e2605822d1626bdebf836c2fb23f2#diff-ad80ececa291657d6a0b9b66ec976607b8e80120134179fba7ca0e46aafa083eR11-R17

I never unconditionally  return `Ok(())`. I return `event::queue::shutdown(exit_code)` in all cases. `event::queue::shutdown(exit_code)` returns `Result` back to WASM guest.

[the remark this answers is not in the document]
no30bit: @cong-or @stevenj 

`ControlFlow` and `HermesEvent` could be split into 2 channels with control flow channel owning the event loop thread.

For example

```rust
/// Hermes event queue.
/// It is a singleton struct.
struct HermesEventQueue {
    /// Hermes event queue sender
    shutdown_receiver: Lock,
    /// Hermes event queue sender
    event_sender: Sender<HermesEvent>,
}

// ...  queue::init() ...
thread::spawn(move || {
    let event_loop = thread::spawn(move || event_execution_loop(& event_receiver));
    let exit = shutdown_receiver.recv();
    exit_tx.set(exit);
    // event loop dropped
});
```

Is that the desired approach?

hermes/bin/src/event/queue/exit.rs
cong-or: ok makes sense

cong-or (approved):
lgtm

hermes/bin/src/app.rs
```diff
@@ -21,7 +21,7 @@ impl std::fmt::Display for ApplicationName {
 }
 
 /// Hermes application
-pub(crate) struct Application {
+pub struct Application {
```
rafal-ch, line 24: Does this (and the `fn load_app()`) need to become `pub`? Maybe it'll be good to double check other places?

hermes/bin/src/event/queue.rs
```diff
@@ -18,6 +23,11 @@ static EVENT_QUEUE_INSTANCE: OnceCell<HermesEventQueue> = OnceCell::new();
 #[error("Failed to add event into the event queue. Event queue is closed.")]
 pub(crate) struct CannotAddEventError;
 
+/// Failed to add event into the event queue. Event queue is closed.
```
rafal-ch, line 26: Possibly a mismatched comment?

wasm/examples/rust/next_century/Earthfile
```diff
@@ -0,0 +1,27 @@
...
+
+IMPORT ../../../wasi AS wasi
+
+# gen-bindings - generates `hermes.rs` bindings to work with.
+gen-bindings:
+    FROM wasi+build-rust-bindings
+
+    SAVE ARTIFACT hermes.rs AS LOCAL src/hermes.rs
+
+# build - builds the http reply integration test wasm component
```
rafal-ch, line 15: Mismatched comment?

wasm/examples/rust/next_century/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+// Allow everything since this is generated code.
+#[allow(clippy::all, unused)]
+mod hermes;
+mod stub;
+
+struct TestComponent;
+
+fn log_shutdown() {
+    const FILE: &str = "next_century/src/lib.rs";
+    const MSG: &str = "Issuing shutdown..";
```
rafal-ch, line 10:
```suggestion
    const MSG: &str = "Issuing shutdown...";
```

rafal-ch (commented):
A couple of minor things/questions.

hermes/bin/src/event/queue.rs
no30bit: Yes, thx

```suggestion
/// Failed to shutdown event queue. Event queue is closed.
```

wasm/examples/rust/next_century/Earthfile
no30bit: Right, thanks for spotting it!

```suggestion
# build - builds the "next_century" Rust wasm component
```

hermes/bin/src/app.rs
no30bit: You mean this is a security and/or an encapsulation concern?

rafal-ch: That makes sense, thanks for explanation.

rafal-ch (approved):
LGTM :+1:

Issue #479, opened 2025-08-07
🛠️ [TASK] : Athena creates/opens an empty SQLite database


rafal-ch:
### Summary

Athena creates/opens an empty SQLite database

### Description

Athena app should create/open an SQLite database on startup.

rafal-ch:
Bumping estimation `5 -> 8` since this also includes the effort to update the athena build system to support separate bindings per module, in order to avoid pulling the entire hermes interface everywhere.

PR #480, opened 2025-08-07
feat(hermes): Athena creates and/or opens an empty SQLite DB on startup

rafal-ch:
# Description

Athena creates and/or opens an empty SQLite DB on startup

## Related Issue(s)

Closes https://github.com/input-output-hk/hermes/issues/479

## Description of Changes

This PR prepares Athena to be able to write to DB. Currently, just some example data is written.
```
$ sqlite3 hermes_datastore.db.app
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
sqlite> .tables
people
sqlite> .headers on
sqlite> select * from people ;
id|name|age
1|Athena|1
```

## Breaking Changes

None.

* [X] I have made corresponding changes to the documentation
* [X] Any dependent changes have been merged and published in downstream module


hermes/apps/athena/modules/db/src/stub.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn on_cardano_immutable_roll_forward(
+        _subscription_id: &hermes::exports::hermes::cardano::event_on_immutable_roll_forward::SubscriptionId,
+        _block: &hermes::exports::hermes::cardano::event_on_immutable_roll_forward::Block,
+    ) {
+    }
+}
+
+impl hermes::exports::hermes::cardano::event_on_block::Guest for DbComponent {
+    fn on_cardano_block(
+        _subscription_id: &hermes::exports::hermes::cardano::event_on_block::SubscriptionId,
```
stevenj, line 29: Don't implement every single possible entry point.  You should only define endpoints that are active.  Hermes should use what you define to determine what events the module needs,  if they are all implemented it is not able to do that.

rafal-ch: The change is implemented, but there are currently two separate `worlds` living in the same `hermes` directory, which is problematic.

Additionally, the modules should be self contained, so the `db.wit` file needs to be moved into the `apps` directory. But for this to work, our build system must be modified. Hence this branch will be on-hold for some time (`no-track`) and additional GH issue will be created.

hermes/apps/athena/modules/Earthfile
```diff
@@ -78,15 +79,35 @@ get-local-hermes:
...
+gen-bindings-http-proxy:
     # Use the WASI build environment which includes WIT binding generation tools
     FROM wasi+build-rust-bindings
 
     # Export generated bindings to the HTTP proxy module source directory
     # This allows the Rust code in http-proxy/src/ to import and use Hermes APIs
-    # NOTE: In multi-module apps, this same hermes.rs is copied to each module's src/
     SAVE ARTIFACT hermes.rs AS LOCAL athena/modules/http-proxy/src/hermes.rs
 
+builder-db:
```
bkioshn, line 90: I think it is better to move this module specific things into its Earthfile so we wont end up having large Earthfile

rafal-ch:
We're not gonna be doing this. Since modules can't speak to each other directly, we can't use a specialized DB module. Every module need to read/write directly to the DB.

Some parts of this PR could be re-used in the `RBAC` module, @bkioshn.

Issue #482, opened 2025-08-08
[Bug]: Broken call to `sqlite3_open_v2()` in the `sqlite` runtime extension


rafal-ch:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The following call:
```
sqlite3_open_v2(
            db_path.to_string_lossy().as_ptr().cast(),
            ...
```
causes the "broken" db_path to be used, for example:
`'hermes_datastore.db'$'\271''mY'` instead of just `hermes_datastore.db`

### Expected Behavior

The file name should be just: `hermes_datastore.db`

### Project

Athena

### Version

It happens on the top of the `main` branch (developer version).

### What operating system are you seeing the problem on?

Linux

### What browsers are you seeing the problem on?

Not Applicable

### Relevant log output

```sh

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #483, opened 2025-08-08
fix(hermes): Correctly pass the db_name to the `sqlite3_open_v2` FFI

rafal-ch:
Closes https://github.com/input-output-hk/hermes/issues/482


minikin (approved):
LGTM

PS: @rafal-ch, it looks like we need to update codeowners.

cong-or (approved):

PR #488, opened 2025-08-12
feat(hermes): Add thread pool for wasm modules parallel execution

### LINKED ISSUE #81: 🛠️ [TASK] : WASM `Module::call_func` implementation parallelisation
### Summary

Parallelise `Module::call_func` function implementation

### Description

Run an internal implementation of the [`Module::call_func`](https://github.com/input-output-hk/hermes/blob/170207ff72c1146e02d6e6c307062fb4f6db439c/hermes/bin/src/wasm/module.rs#L78) in the separate `thread`.
Under the separate `thread` execution should be moved WASM `Store` initialisation, wasm module instantiation, and wasm function call.

```rust
    pub(crate) fn call_func<Args, Ret>(
        &mut self, name: &str, args: Args,
    ) -> Result<Ret, Box<dyn Error>>
    where
        Args: WasmParams,
        Ret: WasmResults,
    {
        self.context.use_for(name.to_string());
        std::thread::spawn(|| {
            let mut store = WasmStore::new(&self.engine, self.context.clone());
            let instance = self.pre_instance.instantiate(&mut store)?;
            let func = instance.get_typed_func(&mut store, name)?;
            Ok(func.call(&mut store, args)?)
        });
    }
```

### Definition of Done

* A test that proves the parallel execution of events

aido-mth:
# Description

This PR adds thread pool, which makes possible parallel execution of wasm modules. Also it provides test, which checks that execution runs in parallel.

## Related Issue(s)

Closes #81

## Description of Changes

1) Add task thread pool (wrapper around `rayon` thread pool)
2) Provided integration test to check if modules actually run in parallel
    - Added component, that sleeps for 1 second per request `tests/integration/components/sleep_component`
    - Added `tests/integration/tests/serial/parallel_module_execution.rs` which can be managed via `cargo test`
3) Updated `hermes/bin/tests/integration/tests/serial/utils/` to be more generic in usage

## Breaking Changes

Describe any breaking changes and the impact.

### DIFF
--- a/hermes/bin/src/pool.rs
+++ b/hermes/bin/src/pool.rs
@@ -12,7 +12,7 @@
 //! concurrently within this pool.
 
 use std::{
-    sync::{atomic::AtomicUsize, Condvar, Mutex, OnceLock},
+    sync::{atomic::AtomicUsize, Condvar, Mutex},
     thread::available_parallelism,
 };
 
@@ -23,12 +23,11 @@ use rayon::ThreadPoolBuilder;
 static TASK_COUNTER: AtomicUsize = AtomicUsize::new(0);
 
 /// Synchronization primitives for waiting until all tasks finish.
-static TASK_WAIT: OnceLock<(Mutex<()>, Condvar)> = OnceLock::new();
+static TASK_WAIT: (Mutex<()>, Condvar) = (Mutex::new(()), Condvar::new());
 
-/// Get a reference to the global `(Mutex, Condvar)` tuple,
-/// initializing it if necessary.
+/// Get a reference to the global `(Mutex, Condvar)` tuple.
 fn get_task_wait() -> &'static (Mutex<()>, Condvar) {
-    TASK_WAIT.get_or_init(|| (Mutex::new(()), Condvar::new()))
+    &TASK_WAIT
 }
 
 /// Initialize the global Rayon thread pool
--- a/.config/dictionaries/project.dic
+++ b/.config/dictionaries/project.dic
@@ -2,8 +2,8 @@ aarch
 aido
 asat
 asyncio
-autocheckpoint
 auditability
+autocheckpoint
 backpressure
 bkioshn
 blockfetch


hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+                },
+                Message::Shutdown => break,
+            }
+        }
+    });
+
+    Worker { handle }
+}
+
+pub(crate) struct Pool {
```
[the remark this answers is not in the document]
aido-mth: It seems to be legit, I'll rewrite everything around `rayon` thread pool.

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+impl Pool {
+    /// Execute a task on the thread pool
+    pub(crate) fn execute(
+        &self,
+        task: Box<dyn FnOnce() -> anyhow::Result<()> + Send + 'static>,
+    ) -> anyhow::Result<()> {
+        if self.terminated.load(Ordering::Acquire) {
+            return Err(anyhow::anyhow!("Thread pool is shut down"));
+        }
+
```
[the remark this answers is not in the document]
rafal-ch: I see this code is now gone, resolving the comment.

hermes/bin/src/app.rs
```diff
@@ -67,59 +68,65 @@ impl Application {
...
-    let runtime_ctx = HermesRuntimeContext::new(
-        app_name,
-        module_id,
-        event.event_name().to_string(),
-        module.exec_counter(),
-        vfs,
-    );
+    event: Arc<dyn HermesEventPayload>,
+) {
+    // TODO: fix how init is processed.
```
stevenj, line 113: In any TODO, add your initials, and a link to an issue that tracks it.  Otherwise the code gets full of TODOs with no way to know who made it, why, and what were they thinking at the time.

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,100 @@
...
+    TASK_WAIT.get_or_init(|| (Mutex::new(()), Condvar::new()))
+}
+
+/// Initialize the global Rayon thread pool
+///
+/// The number of threads is set to the number of available
+/// CPU cores minus two. This subtraction ensures that
+/// the main thread and event loop threads have enough
+/// CPU capacity to handle orchestration and I/O without
+/// being blocked by compute-heavy WASM tasks. At least
```
stevenj, line 40: It won't block them, but it will cause them to have contention which will slow down interactivity (such as serving static assets on http, or managing block chain data).  Maybe the comment could be improved here to not mislead any returning devs to this code.

stevenj (changes requested):

hermes/bin/tests/integration/components/sleep_component/src/stub.rs
```diff
@@ -0,0 +1,83 @@
+use crate::{
```
stevenj, line 1: Have a look at what @rafal-ch is working on, we should not have any unused events inside any wasm module.

hermes/bin/tests/integration/components/sleep_component/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+            .expect("cannot read settings file");
+        let parsed_json: serde_json::Value =
+            serde_json::from_str(&settings_json).expect("unable to parse settings as JSON");
+        let http_server = parsed_json
+            .get("http_server")
+            .expect("missing http_server in settings")
+            .as_str()
+            .expect("http_server is not a string");
+
+        let mut file = std::fs::File::create(RESPONSES_FILE).expect("failed to create file");
```
stevenj, line 46: Its not a great idea to use the filesystem for this because of the issues you noted, but in any event, even using os files this way is problematic without file locking.  I suggest instead use the sqlite DB to keep whatever records, because it will properly handle contention and simultaneous updates/accesses.

stevenj (changes requested):

aido-mth:
Now blocked by [issue](https://github.com/input-output-hk/hermes/issues/517).

rafal-ch:
> Now blocked by [issue](https://github.com/input-output-hk/hermes/issues/517).

Technically, it is not blocked. The parallel execution works and has already been a part of the last sprint demo.

We wanted to improve the testing machinery to use SQL instead of filesystem, but this uncovered some issues in the current SQL implementation and these are now being fixed.

cc: @aido-mth

hermes/bin/src/app.rs
aido-mth: Fixed.

hermes/bin/src/pool.rs
aido-mth: Fixed.

rafal-ch (approved):
LGTM :+1: 
I left two nits.

hermes/bin/src/app.rs
```diff
@@ -79,66 +80,78 @@ impl Application {
...
         for module in self.indexed_modules.values() {
             module_dispatch_event(
-                module,
+                module.clone(),
                 self.name.clone(),
                 module.id().clone(),
                 self.vfs.clone(),
-                event,
-            )?;
+                event.clone(),
```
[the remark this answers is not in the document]
aido-mth: Makes sense, refactored this.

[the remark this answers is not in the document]
aido-mth: > we don't do that anywhere else in the code?

yes, only here

rafal-ch (approved):

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,98 @@
...
+static TASK_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+/// Synchronization primitives for waiting until all tasks finish.
+static TASK_WAIT: OnceLock<(Mutex<()>, Condvar)> = OnceLock::new();
+
+/// Get a reference to the global `(Mutex, Condvar)` tuple,
+/// initializing it if necessary.
+fn get_task_wait() -> &'static (Mutex<()>, Condvar) {
+    TASK_WAIT.get_or_init(|| (Mutex::new(()), Condvar::new()))
+}
```
no30bit, line 32: Isn't `OnceLock`excessive?

```suggestion
/// Synchronization primitives for waiting until all tasks finish.
static TASK_WAIT: (Mutex<()>, Condvar) = (Mutex::new(), Condvar::new());

/// Get a reference to the global `(Mutex, Condvar)` tuple,
/// initializing it if necessary.
fn get_task_wait() -> &'static (Mutex<()>, Condvar) {
    &TASK_WAIT
}
```

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// # Arguments
+///
+/// * `task` - The closure to run concurrently
+pub(crate) fn execute<F>(task: F)
+where F: FnOnce() + Send + 'static {
+    TASK_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+
+    rayon::spawn(move || {
+        task();
```
no30bit, line 68: Doesn't the `TASK_COUNTER` get invalidated if the `task()` panics?
Maybe it should be wrapped into `catch_unwind`.

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/// Wait for all currently submitted tasks to finish
+///
+/// This function blocks until `TASK_COUNTER` reaches zero,
+/// i.e., until all tasks submitted via `execute()` have
+/// completed.
+pub(crate) fn terminate() {
+    let (lock, no_active_tasks_cv) = get_task_wait();
+    if let Ok(mut guard) = lock.lock() {
```
no30bit, line 88: If the `TASK_WAIT` mutex gets poisoned, do we not need to handle it? Say, if the result is `Err` here, shouldn't we terminate immediately instead of no-op?

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,98 @@
...
+/// given task in Rayon’s pool, and decrements the counter
+/// when the task finishes. If it was the last task, it
+/// notifies any threads waiting in `terminate()`.
+///
+/// # Arguments
+///
+/// * `task` - The closure to run concurrently
+pub(crate) fn execute<F>(task: F)
+where F: FnOnce() + Send + 'static {
+    TASK_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
no30bit, line 65: From rustlang [docs](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Acquire)

>Acquire
When coupled with a load, if the loaded value was written by a store operation with [Release](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Release) (or stronger) ordering, then all subsequent operations become ordered after that store. In particular, all subsequent loads will see data written before the store.

So, this should be ordered as `Release` or stronger. Otherwise the later use of `TASK_COUNTER` is `Relaxed`:
```rust
// This is relaxed
let prev = TASK_COUNTER.fetch_sub(1, std::sync::atomic::Ordering::AcqRel); 
```

TL;DR make it `Release` or there's a race.

hermes/bin/src/pool.rs
aido-mth: At this point we don't handle panics. It would be great enhancement in the future.

hermes/bin/src/pool.rs
aido-mth: At this point we don't handle panics. It would be great enhancement in the future.

hermes/bin/src/pool.rs
aido-mth: In this case the fetch_sub is already using Ordering::AcqRel, which covers the Release side needed to synchronize with the Acquire loads in terminate(). That ensures the last decrement “publishes” all prior writes from the task, and the waiter sees them when the counter reaches zero.

The earlier fetch_add(Relaxed) is fine, since it only increments the count before spawning and doesn’t participate in synchronization.

So there isn’t a race here — but if we had used Relaxed on the decrement, then your concern would absolutely apply.

hermes/bin/src/pool.rs
aido-mth: Agreed, this could be simplified

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::{
+    sync::{atomic::AtomicUsize, Condvar, Mutex},
+    thread::available_parallelism,
+};
+
+use anyhow::{Context, Result};
+use rayon::ThreadPoolBuilder;
+
+/// Global counter of currently running tasks.
+static TASK_COUNTER: AtomicUsize = AtomicUsize::new(0);
```
stevenj, line 23: Why is it a usize?  What relationship does it have to pointers or memory?  Is it used for indexing?
Unless its got something to do with memory (an index or pointer math) we should pick an appropriately sized variable type so its 100% consistent regardless of platform.
Given that its actually counting concurrently running threads, it probably never needs to be bigger than u16.

hermes/bin/src/pool.rs
no30bit: I see it now, thanks for the feedback. Either way it seemed a bit convoluted

hermes/bin/src/pool.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// Synchronization primitives for waiting until all tasks finish.
+static TASK_WAIT: (Mutex<()>, Condvar) = (Mutex::new(()), Condvar::new());
```
stevenj, line 26: I feel if the aim here is to build a countdown latch, then we should have a countdown latch type which does that.
Seems like this is the sort of thing we would use in other places.
It would also be easier to test, in isolation.
would also make the intent of it here more obvious.

stevenj (changes requested):

PR #489, opened 2025-08-13
fix(hermes): required wit files live in the DB module

### LINKED ISSUE #487: 🛠️ [TASK] : Move the `db.wit` from `hermes` to the `db` athena module and update the build system to support this
### Summary

Move the `db.wit` from `hermes` to the `db` athena module and update the build system to support this

### Description

This should be built on top the following branch:
https://github.com/input-output-hk/hermes/pull/480

The goal is to move the `db.wit` file from
`wasm/wasi/wit/db.wit`
to
`hermes/apps/athena/modules/db`
because the module should be self-contained and do rely on any files from hermes.

The part of out build system that is responsible for building the bindings will not work with the proposed setup so we need to update it.

rafal-ch:
# Description

This PR changes the way a module is built withing the `athena` app. Currently it only affects the `db` module, since the only other module (`http_proxy`) is under development. We'll adjust the build system for the latter when the functional work is done, to avoid conflicts.

## Related Issue(s)

Closes #487 

## Description of Changes

Previously every module used the entire hermes world to generate bindings. As an effect, every module had to "implement" all possible functions from all interfaces, even the not-used ones, like [here](https://github.com/input-output-hk/hermes/blob/main/hermes/apps/athena/modules/http-proxy/src/stub.rs).

Currently, each module has its own world defined with only the required functions (in this case it's a `db.wit`) and the build system is modified so that this module can "see" the wasi dependencies located in \`hermes\`.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,7 +1,7 @@
 VERSION 0.8
 
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/mdlint:v3.5.4 AS mdlint-ci
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/cspell:v3.5.4 AS cspell-ci
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/mdlint:v3.5.7 AS mdlint-ci
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/cspell:v3.5.7 AS cspell-ci
 
 # cspell: words livedocs sitedocs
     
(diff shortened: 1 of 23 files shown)


hermes/apps/athena/modules/Earthfile
```diff
@@ -86,11 +87,23 @@ gen-bindings-http-proxy:
     # This allows the Rust code in http-proxy/src/ to import and use Hermes APIs
     SAVE ARTIFACT hermes.rs AS LOCAL athena/modules/http-proxy/src/hermes.rs
 
+builder-db:
+    DO rust-ci+SETUP
+
+    COPY wasm+wasi-src/wasi ./wasi
```
rafal-ch, line 93: This should _probably_ copy `hermes` into the `deps` directory, so that `athena` can use `hermes` as a dependency.

PR #491, opened 2025-08-14
fix(general): Bump rust ci to v3.5.6

bkioshn:
# Description
 
Bump rust-ci from v3.5.4 to v3.5.6

This will solve `libc header file not found`  for zstd direct or transitive dependency.


rafal-ch (approved):
LGTM :+1:

Issue #456, opened 2025-07-30
[TASK] Adapt hermes code to the breaking changes in wasmtime 0.35


rafal-ch:
Bumping wasmtime from 0.33 to 0.35 caused compilation errors mainly due to the changes to the `add_to_linker` API.

rafal-ch:
Fixed in https://github.com/input-output-hk/hermes/pull/442

PR #123, opened 2024-02-08
feat: Update wasm::Module to enable wasm component model [try 2]

Mr-Leshiy:
# Description

Updated current wasm::Module implementation to be aligned with the WASM component model design

## Related Issue(s)

Part of https://github.com/input-output-hk/hermes/issues/79

## Description of Changes

- added `event::HermesEventPayload` trait as a Hermes event abstraction.
- updated `wasm::Module` implementation. Made it a WASM component centric. `wasm::Module` incapsulates the logic around "How to load wasm module", "How to prepare a WASM environment for every execution of the wasm module. Also `wasm::Module` does not responsible and does not answering the question "How to execute wasm function", this functionality is generalised with the help of `wasm::module::Instance` trait and `event::HermesEventPayload` trait.

PR #496, opened 2025-08-19
Feat/athena rbac reg

bkioshn:
# Description

Thanks for contributing to the project!
Please fill out this template to help us review your changes.

## Related Issue(s)

> e.g., Closes #123, Resolves #456  Fixes #367

## Description of Changes

## Breaking Changes

Describe any breaking changes and the impact.

## Screenshots

If applicable, add screenshots to help explain your changes.

## Related Pull Requests

If applicable, list any related pull requests.

> e.g., #123, #456


hermes/apps/athena/modules/rbac-registration/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+            hermes::hermes::cardano::api::CardanoNetwork::Preprod => {
+                cardano_blockchain_types::Network::Preprod
+            },
+            hermes::hermes::cardano::api::CardanoNetwork::Preview => {
+                cardano_blockchain_types::Network::Preview
+            },
+            hermes::hermes::cardano::api::CardanoNetwork::TestnetMagic(n) => {
+                // TODO(bkioshn) - This should be mapped to
+                // cardano_blockchain_types::Network::Devnet
+                log_error(
```
stevenj, line 249: we should use https://docs.rs/log/latest/log for logging and a custom implementation of a backend which uses our logging api to do the actual logging.  Not use this directly like this.  This should probably be its own crate, because we will re-use it across all rust modules.   Have a look at this crate which links to lots of example backends for reference.

`file!()` and `line!()` macros etc should be used instead of difficult to maintain hardcoded constant strings.

Don't hard code type names, use `https://doc.rust-lang.org/std/any/fn.type_name.html` or `https://doc.rust-lang.org/std/any/fn.type_name_of_val.html`.  Otherwise if the type changes your log message won't and you have no way to easily detect it.  Using the functions means a change to the type name will prevent compilation until its fixed in the log messages, keeping them accurate.

stevenj (changes requested):

PR #519, opened 2025-08-28
feat(hermes): auto stub untyped

### LINKED ISSUE #502: 🛠️ [TASK] : Investigate the wasi error
### Summary
After we created a separate wasi world for a module (according to [this](https://github.com/input-output-hk/hermes/pull/480#discussion_r2262235780) comment) the system stopped working.

### Description

The system builds correctly but there is an runtime error:
```
{"timestamp":"2025-08-19T11:49:53.194074594Z","level":"ERROR","fields":{"message":"no exported instance named `wasi:http/incoming-handler@0.2.6`"},"target":"hermes::event::queue","filename":"bin/src/event/queue.rs","line_number":115,"threadId":"ThreadId(5)"}
```
as a result, no events are processed.

no30bit:
# Description

Remove `stub.wasm` dependency. Replace with runtime check for export presence. No import-export requirement or other validation yet.

## Related Issue(s)

Closes #502

## Description of Changes

Changes in hermes/bin:
- unchecked_exports.rs  is a fallible mirror to wasmtime export search
RTE/../event.rs files now additionally define their corresponding exports
module.rs Module::init  does not call guest init if it is not exported by user (not an error)
Changes in tests (wasm/example/**, wasm/integration-tests/**, hermes/tests/integration):
- C lang tests: added hermes.wit "specialization" WIT package with hermes:app/hermes . It is used in place of wasm/wasi/wit/hermes.wit, so that only a subset of imports/exports has bindings generated and implemented for
Rust tests: added "inline" hermes:app/hermes world inside of wit-bindgen macro. Same purpouse as hermes.wit for C tests.
- Changes to Earthfiles:
New BUILD_C_COMPONENT & BUILD_RUST_COMPONENT  Earthly functions. They replace boilerplate that was a bit large even prior to this update.
Rust tests in wasm/examples and wasm/integration-tests no longer require [hermes.rs](http://hermes.rs/) bindings generation. It's embedded by wit-bindgen macro call.
C tests still require wit-bindgen  CLI step, but they additionally substitute the root world.
I've left golang test mostly untouched. In my mind, it functions as "stub-module.c" but in GO, testing all the exports.
- Directory wasm/stub-module remains:
It is no longer tied to test execution, however, it itself functions as a test for WIT linking, exporting every function.
- WIT Dependency on wasi:http is removed:
A legacy dependency prior to hermes:http-request implementation. Never used.
It exports wasi:http/incoming-handler handle()  function that is – as far as I know – neither implemented, nor planed to be.

## Breaking Changes

None.

### DIFF
--- a/hermes/bin/src/runtime_extensions/bindings/unchecked_exports.rs
+++ b/hermes/bin/src/runtime_extensions/bindings/unchecked_exports.rs
@@ -53,7 +53,8 @@ macro_rules! define {
         fn $rust_func:ident$(<$l:lifetime>)?($($param_name:ident: $param:ty),* $(,)?) $(-> $return:ty)?;
     )*}) => {::paste::paste! {
         $vis trait $ext_trait {$(
-            #[doc = concat!("Looks up \"", $wit_func , "\" from \"", $wit_interface, "\".\n")]
+            #[doc = concat!("Looks up \"", $wit_func , "\" from \"", $wit_interface, "\".\n\n")]
+            #[doc = "# Panic\n\n Lookup never panics, calling the export can panic if misused (see [`wasmtime::component::Func::call`]).\n"]
             fn [<lookup_ $rust_func>]$(<$l>)?(
                 &self,
                 store: &mut ::wasmtime::Store<$crate::runtime_context::HermesRuntimeContext>
@@ -63,7 +64,8 @@ macro_rules! define {
             >;
 
             #[doc = concat!("Looks up and calls \"", $wit_func , "\" from \"", $wit_interface, "\".\n")]
-            #[doc = concat!("This wraps around [`Self::", stringify!([<lookup_ $rust_func>]), "`].\n")]
+            #[doc = concat!("This wraps around [`Self::", stringify!([<lookup_ $rust_func>]), "`].\n\n")]
+            #[doc = "# Panic\n\n Calling the export can panic if misused (see [`wasmtime::component::Func::call`]).\n"]
             #[allow(unused_parens, dead_code, reason = "needed for codegen in no-return case")]
             fn $rust_func$(<$l>)?(
                 &self,
--- a/hermes/Earthfile
+++ b/hermes/Earthfile
@@ -95,7 +95,7 @@ test-wasm-integration:
     COPY ../wasm/integration-test/localtime+build/localtime.wasm ../wasm/test-components/
     COPY ../wasm/integration-test/logger+build/logger.wasm ../wasm/test-components/
     COPY ../wasm/integration-test/sqlite+build/sqlite.wasm ../wasm/test-components/
-    COPY ../wasm/integration-test/smoke-test+build/smoke-test.wasm ../wasm/test-components/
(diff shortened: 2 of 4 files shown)


hermes/bin/src/runtime_extensions/bindings/unchecked_exports.rs
```diff
@@ -0,0 +1,131 @@
...
+        fn hermes_cardano_event_on_block_on_cardano_block(
+            subscription_id: Resource<SubscriptionId>, block_id: Resource<Block>,
+        );
+
+        #[wit("hermes:cardano/event-on-immutable-roll-forward", "on-immutable-roll-forward")]
+        fn hermes_cardano_event_on_immutable_roll_forward_on_cardano_immutable_roll_forward(
+            subscription_id: Resource<SubscriptionId>, block_id: Resource<Block>,
+        );
+
+        #[wit("hermes:cron/event", "on-cron")]
```
stevenj, line 100: My biggest problem with this is it is not synchronized with the wit files at all, the stub used to be broken if we didn't implement an event defined by the hermes world, will this also refuse to compile if we don't define an implemented event?  Or will it only fail at run time?

no30bit: Could we maintain a stub as a standalone example and `#[test]` against it right here?

I can add a compile time assert inside the macro (with include_str from wit) check too if CI tests aren't enough. But this would impact rust-anayzer performance.

no30bit: Test is agnostic of wit structure (it validates against stub). So, the process could be as follows:

- hermes maintainer adds RTE
- they implement it in stub example app
- the test fails unless they add it to the `define_exports` too
  
 It breaks down to 4 cases (assuming .wit is updated):
  
  | stub example |  macro definition | ci check | user experience 
  | ---- | --- | --- | --- |
  | stub not updated | macro not updated | pass | fail (a bug) |
  | stub updated | macro not updated | fail (unknown export in stub) | unreachable |
  | stub not updated | macro updated | fail (all handler must be exported in stub)| unreachable |
  | stub updated | macro updated | pass | pass |
  
  So there's just 1 case in which maintainer would cause a bug.

stevenj (commented):
I see 4 scenarios that need to be handled, and its not clear to me we do that:
1. Wasm DOES not define an Export (Ok, as long as we don't try and call it). (Runtime extensions need to be able to know it doesn't exist not just fail when calling).
2. Wasm DOES define an Export but its incorrectly defined (wrong parameters, etc) (Checked when module is loaded ONCE, and Not OK, we should refuse to load the module as its invalid).
3. Wasm DOES define an Export and its correct (Checked when module is loaded once, and its OK, so module can load).
4. Any runtime extension can check (at module load) if a module properly implements any export, without calling them. (Conversely also needs to know if any module imports an api, and ideally can call it).

Are all these 4 situation handled or possible here?

bkioshn (commented):
Besides from generating only necessary use function in binding file, the new implementation without stub is really clean ⭐️

PR #520, opened 2025-08-29
feat(hermes): New runtime extension initialization/finalization traits.

stevenj:
# Description

Setup the scaffolding and basic implementation of new traits to allow better runtime extension initialization and finalization.

## Description of Changes

Adds new traits, which if implemented by a runtime extension will get hooked to run initialization or finalization tasks for the runtime extensions, at various points in the lifecycle of the node and an application.

## Breaking Changes

None, all internal.


hermes/bin/src/runtime_extensions/init/executor.rs
```diff
@@ -0,0 +1,102 @@
...
+        init = $init:expr,
+        registry = $registry:ident,
+        rte_trait = $rte_trait:ident,
+        span_label = $span_label:expr,
+        ( $( $arg:expr ),* $(,)? )
+    ) => {{
+        use $crate::runtime_extensions::init::errors::RuntimeExtensionErrors;
+
+        /// Data we use to help sort initialization and finalization of runtime extensions.
+        type InstanceData<'a> = executor::InstanceData<'a, dyn $rte_trait>;
```
no30bit, line 26:
```suggestion
        type InstanceData<'a> = $crate::runtime_extensions::init::executor::InstanceData<'a, dyn $rte_trait>;
```

hermes/bin/src/runtime_extensions/init/errors.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// All individual errors that a runtime extension can make.
+///
+/// Note: ALL Error types MUST include:
+/// * `rte_metadata`, `file`, `line` and `error`.
+///
+/// All other data is customized to match the error.
+#[derive(Clone, thiserror::Error, Debug)]
+#[allow(dead_code)]
+pub(crate) enum RuntimeExtensionError {
```
no30bit, line 146: Wouldn't it be more convenient to have 

```rust
enum RuntimeExtensionErrorKind {
  ResourceAllocation,
  ResourceDeallocation,
  Permission
}

struct RuntimeExtensionError {
   kind: RuntimeExtensionErrorKind,

   /// Metadata about the runtime extension who's resource which failed.
   rte_metadata: RteMetadata,
        
   /* other shared fields */

   /// How much of the resource was available.
   available: String,
        
   /// Resource inner error if any
   #[source]
   error: Option<ThreadSafeError>,
}
```

no30bit: Then you can reduce the `macro_rules` down to `file` and `line` recording and otherwise keep it simpler.

hermes/bin/src/runtime_extensions/init/trait_app.rs
```diff
@@ -0,0 +1,161 @@
...
+    ) -> RteInitResult {
+        Ok(())
+    }
+}
+
+#[traitreg::registry(RteInitApp)]
+static RTE_INIT_APP_REGISTRY: () = ();
+
+/// Locks the Initialized `DashSet` to stop race conditions, but only for specific apps.
+static IS_RTE_APP_LOCK: LazyLock<KeyedLock<String>> = LazyLock::new(KeyedLock::<String>::new);
```
no30bit, line 89: I get that this lock isn't a bottleneck, but it does seem to be implemented quite too primitively.

```rust
pub struct KeyedLock<K: Eq + Hash + Clone>(Mutex<HashMap<K, Arc<Mutex<()>>>>);
```

It also has only `<5k downloads` and has been published a bit over a month ago.

no30bit: Can we instead use the `DashMap`  and do the name-based logic on our side?

stevenj: What i would like is a general purpose key locked dash map.   I started working on one, but to get it right, without any race conditions, is a bigger task than i wanted to tackle here.  I think we should leave this as it is for now, and I will add a ticket to make a general key locked concurrent map (whether it is based on dash map is an implementation detail).  Its a common concurrency problem where you want to have multiple things be able to run, but lock only a subset of them.  The most elegant way is not wrapping dashmap, but either adding key locking to dashmap directly, or fork dashmap.  Wrapping introduces various race conditions (and unnecessary inefficiency) that do not occur if you are doing it at the low levels of the map.

A key locked dashmap would achieve that, so you have operations to read (which are blocking or non blocking) and write which are always blocking but only on the key level.   To be safe it should error if you try and lock another key in the same map (because order of locks can cause deadlocks).  Alternatively it could have a multikey lock.  Again, concurrency issues abound here, but its a massively useful general purpose thing to be able to lock a row for update or even read, so the code around the key is state-fully safe at the key level.

So, for now, we should leave this as is, because it achieves this with little effort.  Without a good technical reason why we shouldn't use it, we should use it as the quickest way to prevent key level locking issues, but agree that we need a key level locked dashmap, which we would then use here to replace both the dashmap and key lock, and we would also be able to use in lots of other places.  Its so useful, it would deserve its own crate.  I could not find any crate that can do this in Rust, which is why i went with this rather than get sidetracked on implementing a properly re-usable key-locked map.

hermes/bin/src/runtime_extensions/init/errors.rs
```diff
@@ -0,0 +1,444 @@
...
+        index: usize,
+    ) -> Option<RuntimeExtensionError> {
+        self.0
+            .get(index)
+            .map(orx_concurrent_vec::ConcurrentElement::cloned)
+    }
+}
+
+/// Thread Safe Error Type
+type ThreadSafeError = Arc<Box<dyn std::error::Error + Send + Sync + 'static>>;
```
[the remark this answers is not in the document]
stevenj: Probably, this arose out of my adventures trying to make MutexPoisoned SEND+SYNC which is impossible as far as i can see.

hermes/bin/src/runtime_extensions/init/trait_runtime.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn fini(self: Box<Self>) -> RteInitResult {
+        Ok(())
+    }
+}
+
+#[traitreg::registry(RteInitRuntime)]
+static RTE_INIT_RUNTIME_REGISTRY: () = ();
+
+/// Is the RTE Initialized
+static IS_RTE_RUNTIME_INITIALIZED: LazyLock<Mutex<bool>> = LazyLock::new(|| Mutex::new(false));
```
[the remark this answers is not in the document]
stevenj: I don't think this works with `fini` which detects if `init()` has actually run and waits if its currently running, but `is_completed()` doesn't do that because it will return `false` if `init()` is running and it won't lock , which the code currently does.  So the status will be `init did not run` which is not true, its running, and so cleanup won't happen.  Seems like it introduces a race condition in `init()` and `fini()`, which we probably won't hit, but would rather not have even theoretically.  The idea being, if `init()` is or has run, then `fini()` gets run reliably.

This is strictly a startup make sure the programmer hasn't done something wrong test,  while using `call_once` might be slightly simpler, i don't think its as robust as what is already there now.

hermes/bin/src/runtime_extensions/init/errors.rs
stevenj: Not exactly this, but based on this suggestion.  Done.

hermes/bin/src/wasm/module.rs
```diff
@@ -98,6 +105,10 @@ impl Module {
             .instantiate_pre(&wasm_module)
             .map_err(|e| anyhow::anyhow!("Bad WASM module:\n {}", e.to_string()))?;
 
+        let id = ModuleId(Ulid::generate());
+
+        RteModule::new().init(app_name, &id)?;
+
         Ok(Self {
             pre_instance,
             engine,
```
bkioshn, line 114: should this id be the same as the id pass to RteModule?

hermes/bin/src/cli/mod.rs
```diff
@@ -11,17 +11,21 @@ use std::{path::PathBuf, process::ExitCode};
...
 };
 
 /// A parameter identifier specifying the log level.
+#[allow(dead_code)]
 const ENV_LOG_LEVEL: &str = "HERMES_LOG_LEVEL";
 
 /// An exit code returned on non-application errors.
 /// This is consistent with `cargo test` return code.
+#[allow(dead_code)]
 const INTERNAL_FAILURE_CODE: u8 = 101;
```
no30bit, line 29:
```suggestion
```

This was something I've introduced to distinguish user failures from node failures. I think it doesn't have to be.

hermes/bin/src/runtime_extensions/init/trait_runtime.rs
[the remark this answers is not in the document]
no30bit: There is [`OnceLock::wait`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.wait) function.
It has been introduced it's already stable in our toolchain.
Wouldn't it work?

hermes/bin/src/runtime_extensions/init/executor.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Generic data for sorting initialization/finalization of runtime extensions.
+pub(crate) struct InstanceData<'a, T: ?Sized + 'static> {
+    /// RTE Registration Data
+    pub register: &'a RegisteredImplWrapper<Box<T>>,
+    /// RTE Instance
+    pub instance: Box<T>,
+}
+
+/// Helper to make running `init` anf `fini` for a Trait registry easy and consistent.
```
bkioshn, line 13: Typo
```suggestion
/// Helper to make running `init` and `fini` for a Trait registry easy and consistent.
```

hermes/bin/src/runtime_extensions/init/errors.rs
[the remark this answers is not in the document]
stevenj: I know why its not SEND+SYNC the point isn't that, the point is there are theoretical errors, which we should never get, but need to handle.  This can catch those.  There is nothing rational the caller can do with it, so its for us to help diagnose if something which should never be possible, actually does happen.  And the way its defined, it catches where in the code its created, so we have a direct link back to where this supposed to be impossible error occurred.  I don't see any value in making a specific PoisonError when the caller has no way to do anything with that.  Its just complicating and obfuscating the error types, which should all be manageable by the caller with a ton of "look at this error, nothing you can do with it, and it shouldn't happen anyway" type errors.  So i am not going to be changing it.  In a future PR if there is some value in disambiguating random shouldn't get this error, and Poison specifically, we can revisit it.  I also want to disuade everyone just throwing every single error they get inside a "GeneralError"   This error is specifically and only to be used for errors which should be impossible under all normal cirucumstances ONLY.  Any other error should end up as one of the defined kinds, or make a new Kind, but if we make a new kind there needs to be some idea of the concrete error procesing attached to it.

For example a failure to allocate "400MB" on disk, will give us some kind of IO error.  The caller DOES NOT CARE what kind of error, its a resource error.  The disk resource WANTED 400MB and could not get it.  That's all the caller cares about.

hermes/bin/src/runtime_extensions/init/trait_runtime.rs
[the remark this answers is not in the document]
stevenj: This is a case of one developer would implement it a different way,  your solution isn't technically better by any measurable yard stick.  It may or may not be doable with `call_once`,  but it doesn't matter.  Its not technically superior, its just "alternate" we need to really avoid the kind of review that goes "I would do it this way, not that way" unless there is some big technical reason eg, "introduction of a new crate, when an existing one we use can suffice"  or "re-inventing the wheel, we have crates/modules which do this",  or "this is technically flawed because <reason>", etc.  No two developers will write the same code, and arguing about whose code is more perfect interferes with us having code which is good enough.  This code is good enough unless you can point out a technical reason its flawed.

hermes/bin/src/runtime_extensions/init/trait_runtime.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// 0 is the default priority, and should be used for most runtime extensions that do
+    /// not care about the order of initialization.
+    fn priority(
+        &self,
+        init: bool,
+    ) -> i32 {
+        RteInitPriority::default().priority(init)
+    }
+
+    /// Initialize the Runtime extension at node startup.
```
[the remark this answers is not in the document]
stevenj: fixed

hermes/bin/src/runtime_extensions/init/executor.rs
stevenj: fixed

hermes/bin/src/runtime_extensions/init/trait_runtime.rs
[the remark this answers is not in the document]
no30bit: I don't argue it's not good enough. 

This is technically flawed because `Mutex` can be initialised the second time to a different value. `OnceLock` cannot by `&self`.
Also, you use `LazyLock` just for its `lock` function, which is arguably not the intended purpose. The intended purpose being running a non-const function in const context (i.e. a static variable).
`LazyLock` is essentially a wrapper over `Once` with said extra feature. The implication of that feature forces you to add the `Mutex`. And since it's too general for the job, you have to add at least 100 lines of code to handle _Should never happen_ cases.

So, one can say it is to some extent re-inventing the wheel.

PR #546, opened 2025-09-16
feat(athena): Handle rollback and roll-forward for block event in `rbac-registration` module

### LINKED ISSUE #542: 🛠️ [TASK] : Rbac registration db indexer - handle roll forward and backward
### Summary

Handle roll forward and backward when exist

### Description

There should be 2 tables
1. Volatile table where it store possible mutable Rbac registration data
2. Persistence table where it store immutable Rbac registration data

### Immutable roll forward
- Immutable roll forward happen when the data in the mutable part become immutable or we can say, the some immutable part is finalized and become immutable.

When the subscription encounter Immutable roll forward event, the following need to be done
1. Data that is going to get purged, need to be added to the Persistence table (become immutable)
2. Purge data backward from a given block - slot in the volatile table

### Immutable roll backward
- Immutable roll backward happen when there is a fork on the chain and some block are no longer valid, so we need to roll back

When the subscription encounter block event with `is_rollback` is true, the following need to be do
1. Purge data forward all blocks after the rollback slot
2. Re-index the data from the rollback slot 

#### Definition of done.
- This is hard to test since Cardano finality take up to 24 hours. Hence, a minimal test can be done by - in the persistence, there should be rbac-registraiton data. For volatile, we can try sending a rbac-registration and see whether it is index into the volatile table or not. After 6 hours - 24 hours. this rbac data should be purge from volatile table and add to the persistence table

bkioshn:
# Description

Handle roll forward and backward in rbac-registration module when exist.

## Related Issue(s)

Closes #542

## Description of Changes

- There should be a tables for both persistent and volatile data where the volatile data need to be stored in memory and persistent need to be stored in disk
- Handle rollback
- Handle roll forward

#### Refactor
Made some refactor to reduce redundancy
- `statement.rs` where it handle all the statement (eg.bind, reset, prepare)
- `query_builder.rs` builder for create a query
- `operation.rs` database operation
- break down insert


stevenj (approved):
LGTM

PR #550, opened 2025-09-18
fix(athena): Move HTTP proxy to dedicated module with self-contained build configuration

cong-or:
FIXES BUILD.

`just build-run-all`

# HTTP Proxy Module Restructuring

This PR reorganizes the HTTP proxy module with a dedicated Earthfile in its own directory.

## Key Changes
- Moved HTTP proxy to standalone directory with independent build config
- Implemented direct WIT binding generation instead of using hermes.rs
- Removed unused stub.rs file and verbose comments
- Updated justfile to reference new module location

## Benefits
- Improved modularity with self-contained modules
- Cleaner codebase and simpler build process
- Better separation of concerns


rafal-ch:
We might just remove the stub completely.

bkioshn (approved):

PR #599, opened 2025-10-24
feat(hermes): Updates to the WASM patcher

### LINKED ISSUE #532: 🛠️ [TASK] : Export instance memory and globals
### Summary

A snapshot of module memory is create at the end of the `init()` function. Each time the module is instantiated again (for example, to handle new event), it's memory is recovered from the snapshot.

Note: Memory **is not** snapshotted after every event, but only after `init()`.

### Description

Each time an event in Hermes is processes a new instance of the WASM module is created. It means that events do not share WASM module state and start "from scratch". Currently there is a workaround for this (events can persist state in DB, send and grab it from external server, etc.) but this is not optimal.

### Original idea
**Exporting Memories and Globals**
* All memories and mutable globals must be host-accessible:
  * Existing exports are preserved.
  * Internal/unexported memories or mutable globals are assigned unique export names to avoid conflicts.
  * Immutable globals do not need exports since their values are constant._

### Definition of done

1. After `init()` is done, a module memory is stored in Hermes (in-mem storage, not a persistent one)

rafal-ch:
# Description

Updates of the WASM parser.

## Related Issue(s)

Closes #532 

## Description of Changes

This PR is a final improvement to the WASM patcher. It improves two things:
* Correctly parses the components with multiple core modules
* Fixes parsing components that contain embedded strings (if there happened to be unmatched `(` or `)` characters inside the string, the parses could get confused)

### DIFF
--- a/hermes/bin/src/wasm/patcher.rs
+++ b/hermes/bin/src/wasm/patcher.rs
@@ -398,6 +398,9 @@ impl Patcher {
             .chars()
         {
             end += 1;
+            // TODO[RC]: We need to be more cautious, since WAT supports both \" escapes in strings
+            // and ; arbitrary comments. See: https://webassembly.github.io/spec/core/text/values.html#strings
+            // Ultimately, this needs to be fixed by using a proper parser.
             if ch == '"' {
                 in_string = !in_string;
             }


aido-mth (dismissed):
LGTM
Just as an idea:
```rust
let mut core_modules_iter = core_modules.into_iter();
let module_0 = core_modules.next();
...;
let other_modules = core_modules_iter.collect();
...;
```

rafal-ch:
> LGTM Just as an idea:
> 
> ```rust
> let mut core_modules_iter = core_modules.into_iter();
> let module_0 = core_modules.next();
> ...;
> let other_modules = core_modules_iter.collect();
> ...;
> ```

Good one :+1: 
Updated in https://github.com/input-output-hk/hermes/pull/599/commits/06535aeefc555fc305764eee3ae6f22f077f85f5

hermes/bin/src/wasm/patcher.rs
```diff
@@ -551,6 +604,7 @@ mod tests {
                 (func $two (;1;) (type 1) (result i32)
                     i32.const 2
                 )
+                )
```
no30bit: A nitpick
```suggestion
            )
```

Issue #595, opened 2025-10-23
[Bug]: `cron` RTE not working as expected


rafal-ch:
After issuing

```rust
   /// Ensure `month`, `day`, `hh` and `mm` specify the time in the future
   let cron_string = bindings::hermes::cron::api::mkcron(None, &[month], &[day], &[hh], &[mm]);
   let cron_tagged = bindings::hermes::cron::api::CronTagged {
       when: cron_string,
       tag: "test-cron-tag".into(),
   };
   bindings::hermes::cron::api::add(&cron_tagged, false);
```

the `on-cron` event is never triggered.

Additionally, when trying to use another API:
```rust
   let result = bindings::hermes::cron::api::delay(500_000_000, "half-a-second".into());
```
we end up getting the following error:

```
thread '<unnamed>' panicked at library/std/src/sys/pal/unix/thread.rs:312:9:
failed to join thread: Resource deadlock avoided (os error 35)
```

PR #633, opened 2025-11-04
fix(hermes): Skip the join in cron queue to prevent a deadlock

rafal-ch:
# Description

This PR fixes the deadlock in the `cron` RTE.

## Related Issue(s)

Closes https://github.com/input-output-hk/hermes/issues/595

## Description of Changes

Due to how the cron extension is designed it is possible that a specific thread will try to join itself, which results in a panic related to deadlock detection. The proposed fix is just a workaround to not join the problematic threads, but ultimately we'd probably need to do some redesign of the `cron` extension.

This PR also provides an integration test checking that the cron events are triggered.


saibatizoku (approved):
Looks good to me.

stevenj (approved):

Issue #703, opened 2025-11-27
🛠️ [TASK] : `.new` topic


bkioshn:
### Summary

`.new` topic

### Description

Implement the `.new` topic for document announcements in the Document Sync Protocol. This covers subscription, broadcasting new doc, keepalive, and timers and retries.

## Task
**1. Subscription**
- Subscribe to `<base>.new` during channel initialization

**2. Boardcasting new document**
- The pre-publish check should be covered in https://github.com/input-output-hk/hermes/issues/630
- Boardcasting doc on `<base>.new` - during `post`

**3. Keep alive mechanism** When there is no new doc to be publish, this need to be triggered

3.1 Quiet-Period Timer
- Maintain per-channel quiet-period timer
- Timer duration: uniform random in [20s, 60s] with jitter
- Reset timer when any .new message is received on the channel
- Rate-limit to avoid sending more than one keepalive per `Tquiet` interval

3.2 Keepalive Publishing
- On timer expire (if no new documents to publish), publish `.new` with `docs=[]`

### Out of scope notes
- Message envelope construction/verification 
- Receiving and processing .new messages

PR #706, opened 2025-12-01
feat(hermes): Add subscription to .new topic

bkioshn:
# Description

Implement subscription to .new topic

## Related Issue(s)

This PR cover 1. and 2.

#703

## Description of Changes

- Subscribe to `<base>.new` during channel initialization
- Boardcast .new `<base>.new` when post is performed (note that the message envelope is not implemented yet)

**Note** 
- Add `FIXME` here https://github.com/input-output-hk/hermes/pull/706/files#diff-735bb0709881f10fb58adb871c4212440a8c0df7b8fcb4db686f3a63d412b09bR124 because I'm unsure about the approach
- Test on multiple nodes is not done yet.

### DIFF
--- a/hermes/apps/athena/modules/doc-sync/src/lib.rs
+++ b/hermes/apps/athena/modules/doc-sync/src/lib.rs
@@ -35,8 +35,8 @@ use shared::{
 /// Doc Sync component - thin wrapper calling host-side implementation.
 struct Component;
 
-/// Doc sync channel name
-const CHANNEL_NAME: &str = "documents";
+/// Default channel name for doc-sync operations
+const DOC_SYNC_CHANNEL: &str = "documents";
 
 impl exports::hermes::init::event::Guest for Component {
     /// Initialize the module.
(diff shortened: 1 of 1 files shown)


hermes/bin/src/runtime_extensions/hermes/doc_sync/host.rs
```diff
@@ -202,18 +199,27 @@ impl HostSyncChannel for HermesRuntimeContext {
...
-        // to a topic before you can publish to it)
-        match self.pubsub_subscribe(topic.clone())? {
-            Ok(_) => tracing::info!("✓ Subscribed to topic: {}", topic),
+        let channel_name = DOC_SYNC_STATE
+            .get(&self_.rep())
+            .ok_or_else(|| wasmtime::Error::msg("Channel not found"))?
+            .value()
+            .clone();
+
+        let topic_new = format!("{channel_name}.new");
```
aido-mth: I don't think we want to mark channel as `new` inside host. Here we should know nothing about ipfs sync mechanism.

aido-mth (changes requested):

hermes/bin/src/runtime_extensions/hermes/doc_sync/host.rs
aido-mth: @bkioshn it might be, but how then you are going to send something to channels like `syn`, `diff`? It would require updating `wit` file to open such channels. So I mean that in this case logic around `new` is more about `athena app logic` then something that any other app developers needs.

PR #791, opened 2026-01-27
fix(hermes): clean stale WASM modules before build

cong-or:
## Summary                                                                                                                                                                                  
  - Add `clean-wasm` recipe that removes stale WASM modules before building                                                                                                                   
  - Only cleans modules older than `shared/Cargo.toml` (wit-bindgen version marker)                                                                                                           
  - Preserves build cache when modules are up-to-date                                                                                                                                         
  - Prevents WASI version mismatch errors (e.g., module expects `@0.2.9`, Hermes provides `@0.2.6`)

### DIFF
--- a/justfile
+++ b/justfile
@@ -301,13 +301,10 @@ run-athena:
 
 # Production build: full web assets, containerized (slow but complete)
 # Use for: final testing, deployments, CI/CD pipelines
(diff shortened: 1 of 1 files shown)


rafal-ch (approved):
LGTM :+1: 
Good one!

